PR #65, opened 2019-08-20
added support for wifi AP mode to class Wifi

squonk11:
added support for wifi AP mode to class Wifi.
In order to use the AP mode the command wifi.connect_to_ap(); must be replaced by wifi.start_softap();. As default only one client is allowed to connect to the AP. If more clients should be allowed wifi.set_softap_max_connections(x); can be used before starting the AP.


PerMalmberg:
@squonk11 Could you please enable changes from maintainers on your PR? 

https://help.github.com/en/articles/allowing-changes-to-a-pull-request-branch-created-from-a-fork#enabling-repository-maintainer-permissions-on-existing-pull-requests

squonk11:
It is already enabled.

PerMalmberg:
It won't let me push to your branch/repo, not authorized even though the instructions above state that's where to push to. Anyway, I pushed my changes to a branch in [my repo](https://github.com/PerMalmberg/Smooth/tree/squonk11-wifiap). It's mainly just C++-ifying the code. I also removed logging the password and instead print the auth mode.

I think a test project for AP-mode is in order before merging. A static web page served by the HTTP server will do.

PerMalmberg:
Weird, now I could push to your PR. Anyhow, have a look.

squonk11:
I just checked your c++ifying - for me it is very interesting to see how this has been done - thank you for that. And of course: these changes are o.k. for me.
I hope that I will be able to setup a simple test project this evening. If not then during the next days.

PerMalmberg:
> thank you for that

Happy to teach :)


> I hope that I will be able to setup a simple test project this evening. If not then during the next days.

Ok, I'll let you have a go at it. I suggest you copy the http_server example and remove things you don't need. Just remember to rename the files and namespace - the folder name must be the same as the namespace the code is in.

squonk11:
Using the http_server example and removing everything not urgently necessary was also my plan. Here I also wanted to remove the `template_data_retriever ` and passing NULL to the `HTTPServerConfig` in this place. Unfortunately this doesn't work - there is an error message during compilation. Is there a way to define a Http Server without using the template engine?
Probably I will not finish with this example today but I hope I can shift it to the PR tomorrow.

PerMalmberg:
If you update your branch from upstream, you can now pass `nullptr` as the `template_data_retriever` argument. It previously required a reference so it didn't work without one being passed in.

squonk11:
now I pushed the test "http_server_ap_test". I had some trouble cross-resolving between local, remote and upstream. I hope now it is o.k.

PerMalmberg:
I made some adjustments:
- No need to specify app_main in the test its generated for you. (how did you build this, I get compilation errors if I leave it in?)
- Renamed the test to access_point as that is what is being tested, not the http-server.

It works nicely :)

I can accept this PR as-is, do you have any additional changed you want to make?

squonk11:
currently I don't have any further changes.
I am building with the build system from esp-idf under Windows since I was having problems with the build system you are using. Here I need to explicitly define a main() function.

PerMalmberg:
> I am building with the build system from esp-idf under Windows since I was having problems with the build system you are using. Here I need to explicitly define a main() function.

Hm, that's odd, especially since I've reworked the build system according to the one IDF uses. You should be able to open the root folder/root CMakeList as a project. Anyway, I'll merge this now. Thank you for your contribution.

squonk11:
Now I understood your question concerning additional changes. I thought this question was related to changes on the support for AP - but probably it was related to additional changes/additions to the Smooth repository?
There are a few things I would need in future:

- support for NVS memory in order to store some key/value pairs (maybe I could try to do this also?)
- support for bluetooth SPP profile (maybe this is not best suited for your library)
- ...

PerMalmberg:
It was mainly about the AP-changes. Please continue discussions re. other items in separate threads.

Issue #68, opened 2019-08-30
Will removing header files and $ENV{IDF_PATH} code from files.cmake cause the Linux build to fail or cause other problems?


KerryRJ:
I have tested this using VS2019 and it works but do not know what the impact to Linux will be. Would a PR be in order?

`set(SMOOTH_LIB_ROOT ${CMAKE_CURRENT_LIST_DIR})

set(SMOOTH_SOURCES
        ${CMAKE_CURRENT_LIST_DIR}/smooth/application/network/http/HTTPServerClient.cpp
        ${CMAKE_CURRENT_LIST_DIR}/smooth/application/network/http/regular/HTTPPacket.cpp
        ${CMAKE_CURRENT_LIST_DIR}/smooth/application/network/http/URLEncoding.cpp
        ${CMAKE_CURRENT_LIST_DIR}/smooth/application/network/http/http_utils.cpp
        ${CMAKE_CURRENT_LIST_DIR}/smooth/application/network/http/HTTPProtocol.cpp
        ${CMAKE_CURRENT_LIST_DIR}/smooth/application/network/http/regular/RegularHTTPProtocol.cpp
        ${CMAKE_CURRENT_LIST_DIR}/smooth/application/network/http/regular/MIMEParser.cpp
        ${CMAKE_CURRENT_LIST_DIR}/smooth/application/network/http/regular/HTTPHeaderDef.cpp
        ${CMAKE_CURRENT_LIST_DIR}/smooth/application/network/http/regular/responses/ErrorResponse.cpp
        ${CMAKE_CURRENT_LIST_DIR}/smooth/application/network/http/regular/responses/FileContentResponse.cpp
        ${CMAKE_CURRENT_LIST_DIR}/smooth/application/network/http/regular/responses/StringResponse.cpp
        ${CMAKE_CURRENT_LIST_DIR}/smooth/application/network/http/regular/responses/HeaderOnlyResponse.cpp
        ${CMAKE_CURRENT_LIST_DIR}/smooth/application/network/http/regular/TemplateProcessor.cpp
        ${CMAKE_CURRENT_LIST_DIR}/smooth/application/network/http/websocket/responses/WSResponse.cpp
        ${CMAKE_CURRENT_LIST_DIR}/smooth/application/network/http/websocket/WebsocketProtocol.cpp
        ${CMAKE_CURRENT_LIST_DIR}/smooth/application/network/http/websocket/WebSocketServer.cpp
[...]

PR #70, opened 2019-08-30
Remove .h and $ENV{IDF_PATH} cpp entries

KerryRJ:
Related to #68


PerMalmberg:
@KerryRJ Why is this needed?

KerryRJ:
Simplify the source files list.
The idf_component_register INCLUDE_DIRS should handle the .h.
I would think that the $ENV{IDF_PATH} .cpp files should be pulled through by the REQUIRES section.

PerMalmberg:
I remember now why I've added the .h files - intellisense/code completion seems to work better (at least in CLion) if they are included.

Also, the two files included via IDF_PATH are needed for when compiling for Linux. So, I'm not accepting this PR. Still, thank you for the effort.

PR #83, opened 2019-09-07
Re-licensing

squonk11:
By this commit I authorize re-licensing of my contributions to the Apache 2.0 license.

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -5,3 +5,4 @@ Below is the list of indiviuals that have contributed to the project.
 |Name|Contribution details
 |----|----|
 |[Per Malmberg](https://github.com/PerMalmberg)|Author and maintainer|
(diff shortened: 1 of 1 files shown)


squonk11:
I was reading the contributing guidelines. There is one thing that astonishes me: you are requesting not to use switch statements. Why is this bad and what is the alternative?

PerMalmberg:
@squonk11 The reason is the possibility of unintentional fall-through from one case to the next (missing `break`), something I've seen happen many times. This rule is one that is often heatedly debated and I am aware it goes against [Core Cpp Guidelines,](https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#es70-prefer-a-switch-statement-to-an-if-statement-when-there-is-a-choice), but so far I've never heard anyone suffer from using `if-else if` instead of a `switch`. The rule has worked well for me for the last two decades. That said, there are always exceptions to the rule so if you happen to have a case where using `switch` clearly is better suited to if-else, I'm willing to listen.

PR #88, opened 2019-09-14
Fix using FD

jeremyjh:
Sorry, not sure how I missed this but native builds are broken because I misspelled FD in the else. I thought I compiled in both configurations but obviously not.


PerMalmberg:
S*it happens :)

PR #93, opened 2019-09-26
Pull container from GitHub registry

PerMalmberg:
Pull container from GitHub registry


PerMalmberg:
Requires auth and it is not recommended to store secrets for workflows while GitHub Actions is in beta.

PR #114, opened 2019-11-05
bugfix and additional functionality for Wifi.cpp 

squonk11:
- bugfix for Wifi::get_mac_address() which now returns the right mac address of the used interface
- added Wifi:: get_local_mac_address(uint8_t m[6]) for getting the binary representation of the mac address
- added Wifi::get_local_ip() for getting the IP address of the ESP32

### DIFF
--- a/lib/smooth/core/network/Wifi.cpp
+++ b/lib/smooth/core/network/Wifi.cpp
@@ -39,7 +39,7 @@ namespace smooth::core::network
 {
     Wifi::Wifi()
     {
-        ip.addr = 0;
+        this->ip.addr = 0;
         tcpip_adapter_init();
         esp_event_handler_register(WIFI_EVENT, ESP_EVENT_ANY_ID, &Wifi::wifi_event_callback, this);
         esp_event_handler_register(IP_EVENT, ESP_EVENT_ANY_ID, &Wifi::wifi_event_callback, this);
@@ -199,51 +199,55 @@ namespace smooth::core::network
     {
         std::stringstream mac;
 
-        uint8_t m[6];
-        
-        wifi_mode_t mode;
-        if(esp_wifi_get_mode(&mode) == ESP_OK) 
+        std::array<uint8_t, 6> m;
+        bool ret = get_local_mac_address(m);
+
+        if (ret)
         {
-            esp_err_t ret;
-            if(mode == WIFI_MODE_STA) 
-                ret = esp_wifi_get_mac(WIFI_IF_STA, m);
-            else
-                ret = esp_wifi_get_mac(WIFI_IF_AP, m);
-
-            if (ret == ESP_OK)
+            for (const auto& v : m)
             {
-                for (const auto& v : m)
+                if (mac.tellp() > 0)
                 {
-                    if (mac.tellp() > 0)
-                    {
-                        mac << "_";
-                    }
-
-                    mac << std::hex << static_cast<int>(v);
+                    mac << "_";
                 }
+
+                mac << std::hex << static_cast<int>(v);
             }
         }
+
         return mac.str();
     }
 
-    esp_err_t Wifi::get_local_mac_address(uint8_t m[6]) const
+    bool Wifi::get_local_mac_address(std::array<uint8_t, 6>& m)
     {
         wifi_mode_t mode;
-        esp_err_t ret;
-        if((ret = esp_wifi_get_mode(&mode)) == ESP_OK)
+        esp_err_t err = esp_wifi_get_mode(&mode);
+
+        if (err == ESP_OK)
         {
-            if(mode == WIFI_MODE_STA) 
-                return esp_wifi_get_mac(WIFI_IF_STA, m);
-            else if(mode == WIFI_MODE_AP)
-                return esp_wifi_get_mac(WIFI_IF_AP, m);
-            else 
-                return ESP_FAIL;
+            if (mode == WIFI_MODE_STA)
+            {
+                err = esp_wifi_get_mac(WIFI_IF_STA, m.data());
+            }
+            else if (mode == WIFI_MODE_AP)
+            {
+                err = esp_wifi_get_mac(WIFI_IF_AP, m.data());
+            }
+            else
+            {
+                err = ESP_FAIL;
+            }
         }
-        
-        return ret;
+
+        if (err != ESP_OK)
+        {
+            Log::error("Wifi", "get_local_mac_address(): {}", esp_err_to_name(err));
+        }
+
+        return err == ESP_OK;
     }
 
-    ip4_addr_t Wifi::get_local_ip() const
+    ip4_addr_t Wifi::get_local_ip()
     {
         return ip;
     }
@@ -281,4 +285,6 @@ namespace smooth::core::network
                                       ip_changed);
         core::ipc::Publisher<network::NetworkStatus>::publish(status);
     }
+
+    ip4_addr_t Wifi::ip;
 }
--- a/lib/smooth/core/network/Wifi.cpp
+++ b/lib/smooth/core/network/Wifi.cpp
@@ -37,11 +37,10 @@ using namespace smooth::core;
 
 namespace smooth::core::network
 {
-    ip4_addr_t Wifi::ip = { 0 };
+    ip4_addr_t Wifi::ip = {0};
 
     Wifi::Wifi()
     {
-        this->ip.addr = 0;
         tcpip_adapter_init();
         esp_event_handler_register(WIFI_EVENT, ESP_EVENT_ANY_ID, &Wifi::wifi_event_callback, this);
         esp_event_handler_register(IP_EVENT, ESP_EVENT_ANY_ID, &Wifi::wifi_event_callback, this);
@@ -249,7 +248,7 @@ namespace smooth::core::network
         return err == ESP_OK;
     }
 
-    // attention: access to this function might have a threading issue.
+    // attention: access to this function might have a threading issue. 
     // It should be called from the main thread only!
     ip4_addr_t Wifi::get_local_ip()
     {
@@ -289,4 +288,5 @@ namespace smooth::core::network
                                       ip_changed);
         core::ipc::Publisher<network::NetworkStatus>::publish(status);
     }
+
 }
--- a/lib/smooth/core/network/Wifi.cpp
+++ b/lib/smooth/core/network/Wifi.cpp
@@ -37,6 +37,8 @@ using namespace smooth::core;
 
 namespace smooth::core::network
 {
+    ip4_addr_t Wifi::ip = {0};
+
     Wifi::Wifi()
     {
         this->ip.addr = 0;
@@ -195,7 +197,7 @@ namespace smooth::core::network
         }
     }
 
-    std::string Wifi::get_mac_address() const
+    std::string Wifi::get_mac_address()
     {
         std::stringstream mac;
 
@@ -247,6 +249,8 @@ namespace smooth::core::network
         return err == ESP_OK;
     }
 
+    // attention: access to this function might have a threading issue. 
+    // It should be called from the main thread only!
     ip4_addr_t Wifi::get_local_ip()
     {
         return ip;
@@ -286,5 +290,4 @@ namespace smooth::core::network
         core::ipc::Publisher<network::NetworkStatus>::publish(status);
     }
 
-    ip4_addr_t Wifi::ip;
 }
--- a/lib/smooth/core/network/Wifi.cpp
+++ b/lib/smooth/core/network/Wifi.cpp
@@ -37,7 +37,7 @@ using namespace smooth::core;
 
 namespace smooth::core::network
 {
-    ip4_addr_t Wifi::ip = {0};
+    ip4_addr_t Wifi::ip = { 0 };
 
     Wifi::Wifi()
     {
@@ -248,7 +248,7 @@ namespace smooth::core::network
         return err == ESP_OK;
     }
 
-    // attention: access to this function might have a threading issue. 
+    // attention: access to this function might have a threading issue.
     // It should be called from the main thread only!
     ip4_addr_t Wifi::get_local_ip()
     {
@@ -288,5 +288,4 @@ namespace smooth::core::network
                                       ip_changed);
         core::ipc::Publisher<network::NetworkStatus>::publish(status);
     }
-
 }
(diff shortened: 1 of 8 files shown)


squonk11:
I created the PR with my modifications. Unfortunately I see that the two checks failed:

1. The formatting check gives the error message that file sizes changed - but why is this an issue?

2. The compilation check complains that `esp_wifi_get_mode()` was not declared. But in my environment it compiles.

lib/smooth/core/network/Wifi.cpp
```diff
@@ -139,10 +141,12 @@ namespace smooth::core::network
             }
             else if (event_id == WIFI_EVENT_AP_START)
             {
+                wifi->ip.addr = 0xC0A80401; // 192.168.4.1
```
PerMalmberg: How about making this a configuration option via menuconfig? Preferably as a string that is parsed if it is possible, or four separate values if need be.

lib/smooth/include/smooth/core/network/Wifi.h
```diff
@@ -73,6 +73,8 @@ namespace smooth::core::network
                                             void* event_data);
 
             [[nodiscard]] std::string get_mac_address() const;
+            esp_err_t get_local_mac_address(uint8_t m[6]) const;
+            ip4_addr_t get_local_ip() const;
```
PerMalmberg: Mark these with [[nodiscard]]
The `m` parameter to `get_local_mac_address` is meant to be an outparameter, but since it is not a reference, does it really work? Oh, it decays to a uint8_t*, doesn't it?

lib/smooth/core/network/Wifi.cpp
```diff
@@ -39,6 +39,7 @@ namespace smooth::core::network
 {
     Wifi::Wifi()
     {
+        ip.addr = 0;
```
PerMalmberg: do `ip{};` in the declaration in the header file instead to zero-initialize the entire struct.

lib/smooth/core/network/Wifi.cpp
```diff
@@ -194,23 +200,54 @@ namespace smooth::core::network
...
+        
+        wifi_mode_t mode;
+        if(esp_wifi_get_mode(&mode) == ESP_OK) 
         {
-            for (const auto& v : m)
+            esp_err_t ret;
+            if(mode == WIFI_MODE_STA) 
+                ret = esp_wifi_get_mac(WIFI_IF_STA, m);
+            else
+                ret = esp_wifi_get_mac(WIFI_IF_AP, m);
```
PerMalmberg: More succinct: `esp_err_t ret = esp_wifi_get_mac(mode == WIFI_MODE_STA ? WIFI_IF_STA: WIFI_IF_AP, m);`

Also, if-else are missing braces which uncrustify will complain about.

lib/smooth/core/network/Wifi.cpp
```diff
@@ -194,23 +200,54 @@ namespace smooth::core::network
...
-                    mac << "_";
-                }
+                    if (mac.tellp() > 0)
+                    {
+                        mac << "_";
+                    }
 
-                mac << std::hex << static_cast<int>(v);
+                    mac << std::hex << static_cast<int>(v);
+                }
```
PerMalmberg: Have a look if you can use std::fmt instead of a stringsteam to build the string-version of the mac-address. I believe it is less stack-intense.

lib/smooth/include/smooth/core/network/Wifi.h
```diff
@@ -73,6 +73,8 @@ namespace smooth::core::network
                                             void* event_data);
 
             [[nodiscard]] std::string get_mac_address() const;
+            esp_err_t get_local_mac_address(uint8_t m[6]) const;
```
PerMalmberg: Lets return a bool instead, we succeed or we don't. Unless you actually need know the underlying reason?

I try to keep IDF-types out of the API if possible.

lib/smooth/include/smooth/core/network/Wifi.h
PerMalmberg: Just a thought, smooth has an IP4-class, could it be used as a return value?

lib/smooth/core/network/Wifi.cpp
```diff
@@ -194,23 +200,54 @@ namespace smooth::core::network
...
         }
-
         return mac.str();
     }
 
+    esp_err_t Wifi::get_local_mac_address(uint8_t m[6]) const
+    {
+        wifi_mode_t mode;
+        esp_err_t ret;
+        if((ret = esp_wifi_get_mode(&mode)) == ESP_OK)
```
PerMalmberg: No side effects in if-statements please. Move the assignment out of the if.

lib/smooth/core/network/Wifi.cpp
```diff
@@ -194,23 +200,54 @@ namespace smooth::core::network
...
+        {
+            if(mode == WIFI_MODE_STA) 
+                return esp_wifi_get_mac(WIFI_IF_STA, m);
+            else if(mode == WIFI_MODE_AP)
+                return esp_wifi_get_mac(WIFI_IF_AP, m);
+            else 
+                return ESP_FAIL;
```
PerMalmberg: No multiple returns.
Also, this is essentially the same code as above, perhaps refactor to a separate function?

PerMalmberg:
Also, please update the contributors.md file to list this PR :+1:

squonk11:
thank you for reviewing the code. I will take care about your suggestions tomorrow.

lib/smooth/core/network/Wifi.cpp
squonk11: I am not sure if it makes sense to make it a configuration option because here the IP address of the interface is not being set; it is set internally in esp-idf to 192.168.4.1 when configured as AP. I did not find any way to read the IP address from esp-idf when the AP mode is started. So, I just set the internal member variable of Wifi to this value (hoping that in future versions of esp-idf this address will not be changed)

lib/smooth/include/smooth/core/network/Wifi.h
squonk11: yes actually it is uint8_t*
I was hoping that defining it like this makes the compiler check the amount of data written to the array (but probably this is not the case).

lib/smooth/core/network/Wifi.cpp
squonk11: the struct `ip_addr_t` defined in "ip4_addr.h" contains only the ip address as a `u32_t` - nothing else. So, there are no other members to initialize.

lib/smooth/core/network/Wifi.cpp
squonk11: o.k.

lib/smooth/core/network/Wifi.cpp
squonk11: AFAIK, this is code originally written by you. But I can try to modify it accordingly.

lib/smooth/include/smooth/core/network/Wifi.h
squonk11: o.k.

lib/smooth/include/smooth/core/network/Wifi.h
squonk11: I only need the ip address as `uint32_t` or `uint8_t[4]`. The IP4 class additionally includes port number and other stuff which I don't need and which in the case of UDP and/or in the Wifi class does not make so much sense. In case of a TCP/IP socket connection these information makes more sense - so there could be a member function of the Socket class for getting such information?

lib/smooth/core/network/Wifi.cpp
squonk11: o.k.

lib/smooth/core/network/Wifi.cpp
squonk11: o.k.

lib/smooth/include/smooth/core/network/Wifi.h
PerMalmberg: Fair point.

Btw, I think these two methods should be static; you're unlikely to have access to the Wifi-instance where you need them and I don't want to have to pass it around just to get the IP/mac.

lib/smooth/core/network/Wifi.cpp
PerMalmberg: Ha, right  you are. I'll have a look at it :)

lib/smooth/core/network/Wifi.cpp
PerMalmberg: You're technically right and even though it is unlikely to change, doing ip{} ensures that if it ever changes the entire struct is still zero-initialized.

lib/smooth/include/smooth/core/network/Wifi.h
PerMalmberg: Depending on what error values are originally returned, you could add some error logging if it fails.

lib/smooth/core/network/Wifi.cpp
squonk11: o.k.

lib/smooth/include/smooth/core/network/Wifi.h
PerMalmberg: Well, you could use `std::array` and pass the pointer returned by data() to the idf-function. At least that adds safety upwards and prevents anything other than the correct size from being used.

lib/smooth/core/network/Wifi.cpp
PerMalmberg: Ooh, I see. So this is just to keep it in synch with IDF.

Is there really no way to control what IP the AP gets?

lib/smooth/include/smooth/core/network/Wifi.h
squonk11: Isn't this valid for `std::string get_mac_address() const;` also?

lib/smooth/core/network/Wifi.cpp
squonk11: When using for wifi the [TCP/IP Adapter library](https://docs.espressif.com/projects/esp-idf/en/latest/api-reference/network/tcpip_adapter.html) the IP and MAC can be set using `tcpip_adapter_ap_start.` When using the [Wi-Fi API](https://docs.espressif.com/projects/esp-idf/en/latest/api-reference/network/esp_wifi.html) I did not find a function call to set the IP. 
Since you started the Smooth library using the Wi-Fi API I just continued using it. Maybe it makes sense to switch to the TCP/IP Adapter library later?

lib/smooth/core/network/Wifi.cpp
squonk11: if I define `get_local_mac_address` as static I also have to define `ip` static in order to access it. If `ip` is static the compiler complains:

> error: 'constexpr' needed for in-class initialization of static data member 'ip4_addr_t smooth::core::network::Wifi::ip' of non-integral type [-fpermissive]

when I want to initialize it with ip{}. So, I let the initialization in the Wifi ctor for the moment unless you have another suggestion.

lib/smooth/include/smooth/core/network/Wifi.h
squonk11: o.k. now using `std::array`.

lib/smooth/core/network/Wifi.cpp
PerMalmberg: Yeah, doing that merits its own PR.

lib/smooth/core/network/Wifi.cpp
PerMalmberg: Is the call to the esp_ function costly? If not then ìp` can be put into the function that uses it and just return it by value.

squonk11:
Now I saw that you last comments were in parallel with my last commit. Maybe you can check the actual status and give additional comments on that?

lib/smooth/include/smooth/core/network/Wifi.h
```diff
@@ -89,5 +93,6 @@ namespace smooth::core::network
             std::string ssid{};
 
             std::string password{};
+            static ip4_addr_t ip;
```
PerMalmberg: `ip` can now be accessed without ever creating an instance of `Wifi` which is the only one that initializes it meaning get_local_ip() can return uninitialized data.

`ip` must be initialized out-of-line in the top of Wifi.cpp: `ip4_addr_t Wifi;;ip = {0};`

Do we have threading issue here? `Wifi` is owned by the main task, but Wifi::get_local_ip()` may be called from any thread.

squonk11: I moved the initialization (`ip4_addr_t Wifi::ip = {0};`) to the top of Wifi.cpp.
Concerning threading issue: unfortunately I am not very familiar with that. Sure, here (at least in theory) the main thread and the other thread could access Wifi::ip simultaneously. In my application I am currently calling `get_local_ip()` from the main thread only - so, I am safe. But for general use it would be more safe to make it thread-safe. I never did such things. How it can be realized?

PerMalmberg: We'd need to put a mutex-lock where we read and write it. I'm not very keen on adding a mutex for a single value, the overhead is so large. Let me think about it a while, we can put a warning comment on the get_local_ip function for now.

squonk11: I will do this during the next days (hopefully tomorrow evening).

lib/smooth/core/network/Wifi.cpp
```diff
@@ -37,8 +37,11 @@ using namespace smooth::core;
 
 namespace smooth::core::network
 {
+    ip4_addr_t Wifi::ip = { 0 };
+
     Wifi::Wifi()
     {
+        this->ip.addr = 0;
```
PerMalmberg: This line is no longer needed since it is initialized just above.

squonk11: ups, I missed that.

PerMalmberg:
Are you planning to update a test project to use this new functionality?

squonk11:
Yes, I could modify the AP test.

Am 7. November 2019 20:33:55 MEZ schrieb Per Malmberg <notifications@github.com>:
>Are you planning to update a test project to use this new
>functionality?
>
>-- 
>You are receiving this because you authored the thread.
>Reply to this email directly or view it on GitHub:
>https://github.com/PerMalmberg/Smooth/pull/114#issuecomment-551230173

-- 
Diese Nachricht wurde von meinem Android-Gerät mit K-9 Mail gesendet.

PerMalmberg:
> Yes, I could modify the AP test

Cool, lets finish with that before merging.

squonk11:
I try to do it this evening.

PerMalmberg:
I'm going to accept this as-is now, but will make a few modifications afterwards as I can't push to the  master branch of your repository.

PR #116, opened 2019-11-08
Complete SPI API, added display ILI9341, added spi bme280, updated i2c bme280

enelson1001:
Also added an spi test and an i2c test in test folder but did not update CMakeLists,txt


lib/smooth/application/io/i2c/BME280.cpp
```diff
@@ -14,15 +14,17 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
...
-#include <smooth/core/util/FixedBuffer.h>
-#include <smooth/core/util/ByteSet.h>
+#include <smooth/core/logging/log.h>
+
+using namespace smooth::core;
+using namespace smooth::core::logging;
 
 namespace smooth::application::sensor
 {
+    static const char TAG[] = "BME280SPI:";
```
PerMalmberg: I believe the ":" is redundant here?

Nitpicking - prefer char*  over [] like in the rest of the code base.

lib/smooth/application/io/i2c/BME280.cpp
```diff
@@ -59,217 +61,118 @@ namespace smooth::application::sensor
...
-        core::util::FixedBuffer<uint8_t, 8> measurement{};
+        core::util::FixedBuffer<uint8_t, 8> measurement_data{};
 
         humidity = 0;
         pressure = 0;
         temperature = 0;
 
         bool res = read_trimming_parameters();
-        res = res && read(address, PRESS_MSB_REG, measurement);
+        res = res && read(address, BME280Core::PRESS_MSB_REG, measurement_data);
```
PerMalmberg: Seeing as these two assignments to res are beside eachother I think we can simplify this to just 
```
res = read(...) 
    && read(..)
```

lib/smooth/application/io/spi/BME280SPI.cpp
```diff
@@ -0,0 +1,256 @@
...
+                                       BME280Core::OverSampling& pressure,
+                                       BME280Core::OverSampling& temperature,
+                                       BME280Core::StandbyTimeMS& standby,
+                                       BME280Core::FilterCoeff& filter,
+                                       BME280Core::SpiInterface& spi_interface)
+    {
+        // The spi rx buffer need to be length of multiples of 32 bits to avoid heap corruption.c
+        alignas(4) core::util::FixedBuffer<uint8_t, 4> ctrl_hum_data;
+        alignas(4) core::util::FixedBuffer<uint8_t, 4> ctrl_meas_data;
+        alignas(4) core::util::FixedBuffer<uint8_t, 4> config_data;
```
PerMalmberg: You're lucky this compiles, considering there's no support for alignas or std::aligned_storage as per [this issue](https://github.com/espressif/esp-idf/issues/4218).

If the size is important, I think a static_assert is in order to ensure that the size is actually a multiple of 32 bits.

lib/smooth/application/io/spi/BME280SPI.cpp
```diff
@@ -0,0 +1,256 @@
...
+
+    bool BME280SPI::read_configuration(BME280Core::SensorMode& mode,
+                                       BME280Core::OverSampling& humidity,
+                                       BME280Core::OverSampling& pressure,
+                                       BME280Core::OverSampling& temperature,
+                                       BME280Core::StandbyTimeMS& standby,
+                                       BME280Core::FilterCoeff& filter,
+                                       BME280Core::SpiInterface& spi_interface)
+    {
+        // The spi rx buffer need to be length of multiples of 32 bits to avoid heap corruption.c
```
PerMalmberg: Trailing "c".

Is there a reference to point to re. the risk of heap corruption?

lib/smooth/application/io/spi/BME280SPI.cpp
```diff
@@ -0,0 +1,256 @@
...
+    {
+        if (!trimming_read)
+        {
+            // The spi rx buffer need to be length of multiples of 32 bits to avoid heap corruption.
+            alignas(4) core::util::FixedBuffer<uint8_t, 32> calib00_calib25_data;    // 0x88-0xA1
+            alignas(4) core::util::FixedBuffer<uint8_t, 8> calib26_calib32_data;     // 0xE1-0xE7
+            core::util::FixedBuffer<uint8_t, 32> calibration_data;
+
+            trimming_read = read(BME280Core::CALIB00_REG, calib00_calib25_data.data(), 27) &&
+                            read(BME280Core::CALIB26_REG, calib26_calib32_data.data(), 8);
```
PerMalmberg: `&&` at [beginning of lines](https://github.com/PerMalmberg/Smooth/blob/master/CONTRIBUTING.md).

lib/smooth/application/io/spi/BME280SPI.cpp
```diff
@@ -0,0 +1,256 @@
...
+
+            if (trimming_read)
+            {
+                // don't copy dummy byte but copy 0x88-0x9F into calibration_data
```
PerMalmberg: I admit I haven't read the SPI spec for BME280 - what's the dummy byte about? Can you add a short explanation about it for future reference?

lib/smooth/core/io/spi/SPIDevice.cpp
```diff
@@ -105,6 +106,50 @@ namespace smooth::core::io::spi
...
+            Log::error(log_tag, "polling_write() failed");
+        }
+
+        return res == ESP_OK;
+    }
+
+    bool SPIDevice::queue_transaction(spi_transaction_t& transaction)
+    {
+        // Attach ourselves as the user-data
+        transaction.user = this;
```
PerMalmberg: Are there any threading issues doing this? What if a SPI device is destroyed while a transmission is in progress?

lib/smooth/include/smooth/application/display/ILI9341.h
```diff
@@ -0,0 +1,162 @@
...
+   following: IM0=0, IM1=1, IM2=1, IM3=1.  This jumper setting configures the
+   ILI9341 to operates as a 4-wire 8-bit data serial interface II see page 26
+   and 38 of datasheet for more details.
+
+   NOTE 2: This driver implements both reading and writing to display.  Read is used
+   for reading parameters from certain registers.  The maximum spi clock speed is
+   16MHz but data sheet recommends 10MHz.
+
+   NOTE 3: Most applications will not need the ability to read regsiters and are
+   only writing to the display.  In this case MISO can be set to -1.  Also the spi
```
PerMalmberg: There is actually an enum to use as not connected: GPIO_NUM_NC
https://github.com/espressif/esp-idf/blob/a45e9985344575a80acb1dc9c12e7bec4d8af401/components/driver/include/driver/gpio.h#L242

lib/smooth/include/smooth/application/io/spi/BME280Core.h
```diff
@@ -0,0 +1,233 @@
...
+            enum SensorMode
+            {
+                Sleep = 0x00,
+                Forced = 0x01,
+                Normal = 0x03
+            };
+
+            static const constexpr char* SensorModeStrings[] = { "Sleep",
+                                                                 "Forced",
+                                                                 "Forced",
```
PerMalmberg: Two "Forced"?

lib/smooth/include/smooth/application/io/spi/BME280Core.h
```diff
@@ -0,0 +1,233 @@
...
+            /// Populate the trimming registers for later use when calculating the compensated temp, press and hum
+            // values
+            /// \param calibration_data A container that holds the data read from calibration registers 0x88..0x9F,
+            // 0xA1, 0xE1..0xE7.
+            void populate_trimming_registers(core::util::FixedBuffer<uint8_t, 32>& calibration_data);
+
+        private:
+            // These calculation methods are based on those in the datasheet at
+            // https://ae-bst.resource.bosch.com/media/_tech/media/datasheets/BST-BME280_DS001-11.pdf
+            // as of 2017-08-20. The revision of the code is 1.1 according to that document.
```
PerMalmberg: This comment should be moved to where the implementation is.

lib/smooth/include/smooth/application/io/spi/BME280Core.h
```diff
@@ -0,0 +1,233 @@
...
+            typedef int32_t BME280_S32_t;
+            typedef uint32_t BME280_U32_t;
+            typedef int64_t BME280_S64_t;
```
PerMalmberg: Perhaps originating from my SPI-code, but lets use C++17's using instead of typedef for these three.

example: `using BME280_S32_t = int32_t `

lib/smooth/application/io/spi/BME280SPI.cpp
```diff
@@ -94,7 +94,12 @@ namespace smooth::application::sensor
 
     uint8_t BME280SPI::read_id()
     {
-        alignas(4) std::array<uint8_t, 4> rxdata;
+        std::array<uint8_t, 4> rxdata;
+
+        // The spi rx buffer need to be length of multiples of 32 bits to avoid heap corruption.
+        // see esp-idf/components/driver/spi_master.c line 865
+        //static_assert(&rxdata % 4 == 0, "Error:c rxdata not 32 bit aligned ");
+
```
PerMalmberg: You can do it like this:
```
using BuffType = std::array<uint8_t, 4>;
BuffType foo{};
static_assert(sizeof(BuffType::value_type) * foo.size() % 4 == 0, "Buffer must have a length of multiple of 32 bits, i.e. 4 bytes");
```

You can then place the static assert at a single place where BuffType is used to declare a member. Just make sure to name the type better than my example :)

enelson1001:
**From esp32 API- spi master**

>Normally, the data that needs to be transferred to or from a Device will
be read from or written to a >chunk of memory indicated by the members
rx_buffer and tx_buffer of the structure spi_transaction_t
<https://docs.espressif.com/projects/esp-idf/en/latest/api-reference/peripherals/spi_master.html#_CPPv417spi_transaction_t>.
If DMA is >enabled for transfers, the buffers are required to be:


   1. Allocated in DMA-capable internal memory. If external PSRAM is enabled
   <https://docs.espressif.com/projects/esp-idf/en/latest/api-reference/system/mem_alloc.html#dma-capable-memory>,
   this means using pvPortMallocCaps(size, MALLOC_CAP_DMA).
   2. 32-bit aligned (staring from a 32-bit boundary and having a length of
   multiples of 4 bytes).

>If these requirements are not satisfied, the transaction efficiency will
be affected due to the >allocation and copying of temporary buffers.


Since we are using the SPI device with DMA I would like to make a class
like your FixedBuffer maybe call it DMAFixedBuffer that would use
heaps_caps_malloc(size, MALLOC_CAP_DMA).  I would make this class do static
assert to make sure size is modulus of 4.  I think I would need to
implement free()? So this will take me a few days to figure out. This type
of buffer (DMAFixedBuffer) would be used when the SPI device used dma.
Does this sound like a good approach?

PerMalmberg:
I does sound like a good idea. Re. free(), there should be a heap_caps_free or similar for you to use.

I suggest you do it RAII-style, i.e. allocate the buffer in the constructor and deallocate in the destructor so that the user doesn't have to think about memory allocations.

I do wonder how to ensure that the buffer is aligned on a 32-bit boundary since alignas() isn't yet supported. This should be done in the SPI master code (the temporary buffers the text mentions), but I can't see it.

lib/smooth/include/smooth/core/util/DmaFixedBuffer.h
```diff
@@ -0,0 +1,89 @@
...
+    class DmaFixedBuffer
+    {
+        public:
+            DmaFixedBuffer()
+            {
+                buff = static_cast<T*>(heap_caps_malloc(Size, MALLOC_CAP_DMA));
+
+                // Force the user to create DmaFixedBuffer in multiples of 4 bytes
+                static_assert(sizeof(T) * Size % 4 == 0,
+                              "DmaFixedBuffer must have a length of multiple of 32 bits, i.e. 4 bytes");
```
PerMalmberg: Two things:
- This is evaluated at compile-time so mixing it with the run-time code just confuses things imho. I prefer to place them at the top of the class.
- While the multiples of 4 bytes applies to your specific use case, lets not make this class only usable for this particular usage, shall we? I suggest you do one of these just below `public:`:

```
constexpr size_t size() { return sizeof(T) * Size; }
```
That way it is possible to do a static assert on the buffer instance itself where it is used: `static_assert(buffer_instance::size() % 4 == 0, "blah blah")`

(I hope I got the syntax correct)

lib/smooth/include/smooth/core/util/DmaFixedBuffer.h
```diff
@@ -0,0 +1,89 @@
...
+                    smooth::core::logging::Log::error("DmaFixedBuffer:", "Failed to allocate DmaFixedBuffer");
+                }
+
+                // Inform user that DmaFixedBuffer address is not 32 bit aligned.  Currently cannot use alignas
+                // specifier. The program will work but will be less efficient
+                if (reinterpret_cast<int>(&buff) % 4 != 0)
+                {
+                    smooth::core::logging::Log::warning("DmaFixedBuffer:",
+                                                        "Not 32 bit alligned, spi transaction efficiency will be reduced");
+                }
```
PerMalmberg: Has the user actually any control over what address the buffer ends up at?

I think a better solution is provide the buffer with a constructor argument that specifies the memory type that is needed, i.e. [MEMORY_CAP_8_BIT and MEMORY_CAP_32BIT](https://docs.espressif.com/projects/esp-idf/en/latest/api-reference/system/mem_alloc.html#c.MALLOC_CAP_8BIT) so the end result is `heap_caps_malloc(Size, MALLOC_CAP_DMA | other_type)`. this would users of this class to also add MALLOC_CAP_INTERNAL if needed.

lib/smooth/include/smooth/core/util/DmaFixedBuffer.h
```diff
@@ -0,0 +1,89 @@
...
+            DmaFixedBuffer()
+            {
+                buff = static_cast<T*>(heap_caps_malloc(Size, MALLOC_CAP_DMA));
+
+                // Force the user to create DmaFixedBuffer in multiples of 4 bytes
+                static_assert(sizeof(T) * Size % 4 == 0,
+                              "DmaFixedBuffer must have a length of multiple of 32 bits, i.e. 4 bytes");
+
+                // Inform user if problems with allocating heap for this DmaFixedBuffer
+                if (buff == NULL)
```
PerMalmberg: Doesn't heap_caps_malloc() call abort() if it fails to allocate memory? Perhaps that is just when C++ exceptions are enabled and new() fails? Worth verifying at least since if it is true, the null check doesn't make sense. Also, prefer `nullptr`. https://en.cppreference.com/w/cpp/language/nullptr

lib/smooth/include/smooth/core/util/DmaFixedBuffer.h
```diff
@@ -0,0 +1,89 @@
...
+            }
+
+            [[nodiscard]] size_t size()
+            {
+                return Size;
+            }
+
+            T* data()
+            {
+                //return &buff[0];
```
PerMalmberg: Leftover.

enelson1001:
OK will work on that.

enelson1001:
Yes sounds better, will work on it.

enelson1001:
heap_caps_malloc returns pointer to memory allocated on success and NULL on
failure.
But you would like this removed from class and have the class that uses
DmaFixedBuffer to do this test.  Is that correct?

lib/smooth/include/smooth/core/util/DmaFixedBuffer.h
```diff
@@ -28,26 +22,29 @@ limitations under the License.
 
 namespace smooth::core::util
 {
-    template<typename T, size_t Size>
+    template<typename T, size_t Size, uint32_t MallocCapType>
```
PerMalmberg: The way templates work, having the memory capabilities as a template parameter results in multiple types if different capabilities are used. This in turn results in large code size, which is a scarce resource on embedded system. I suggest passing this as a constructor argument instead.

lib/smooth/include/smooth/core/util/DmaFixedBuffer.h
```diff
@@ -28,26 +22,29 @@ limitations under the License.
...
 
-                // Inform user that DmaFixedBuffer address is not 32 bit aligned.  Currently cannot use alignas
-                // specifier. The program will work but will be less efficient
+                // The user has no control were the compiler will place the DmaFixedBuffer at since we can't use
+                // alignas to force compiler to place DmaFixedBuffer on a 32 bit aligned address.  But it would
+                // be nice to inform user that DmaFixedBuffer address is not 32 bit aligned and that the efficiency
+                // of the program will suffer since extra coping will be required.
                 if (reinterpret_cast<int>(&buff) % 4 != 0)
                 {
                     smooth::core::logging::Log::warning("DmaFixedBuffer:",
```
PerMalmberg: Is this really true? I've not seen (or at least remember) alignment being a requirement, only that the size is a multiple of 32bits which is a different thing.

enelson1001:
Is this really true? I've not seen (or at least remember) alignment being a
requirement, only that the size is a multiple of 32bits which is a
different thing.

This was requirement in SPI Master -- few emails back were you commented on
it.

enelson1001:
I probably should delete the following from DmaFixedBuffer

```C
// Inform user if problems with allocating heap for this DmaFixedBuffer
if (buff == nullptr)
{
    smooth::core::logging::Log::error("DmaFixedBuffer:", "Failed to
allocate DmaFixedBuffer");
}

// The user has no control were the compiler will place the DmaFixedBuffer
at since we can't use
// alignas to force compiler to place DmaFixedBuffer on a 32 bit aligned
address.  But it would
// be nice to inform user that DmaFixedBuffer address is not 32 bit aligned
and that the efficiency
// of the program will suffer since extra coping will be required.
if (reinterpret_cast<int>(&buff) % 4 != 0)
{
    smooth::core::logging::Log::warning("DmaFixedBuffer:",
                                        "Not 32 bit alligned, spi
transaction efficiency will be reduced");
}
```
But it would be nice if I could encapsulate the creation of buffer,
checking buffer size is multiple of 4 bytes, checking address to see if 32
bit aligned and checking if buffer was allocated. I create a lot of
DmaFixedBuffers in BME280SPI and would be nice not to have to repeat the
following code multiple times.

```C
// Since we chose to use DMA on this SPI Device use DmaFixedBuffer for the
rx_buffer in spi_transfer
core::util::DmaFixedBuffer<uint8_t, 16> rxdata(MALLOC_CAP_DMA);

// The SPI Master requires the DMA data to be length of multiples of 32
bits to avoid heap corruption.
static_assert(rxdata.size() % 4 == 0, "rxdata must have a length of
multiple of 32 bits, i.e. 4 bytes");

// Check if rxdata address is 32 bit aligned
if (reinterpret_cast<int>(&rxdata) % 4 != 0)
{
    Log::warning(TAG, "rxdatais not 32 bit alligned, spi transaction
efficiency will be reduced");
}

// Check if buffer was allocated
bool res = rxdata.data() != nullptr;

if(res)
{
    now do stuff with DmaFixedBuffer
}
```

I thought about including additional code in the DMAFixedBuffer such as:
is_buffer_allocated()
is_buffer_address_32_bit_aligned()
is_buffer_size_valid(modulus_value)
but didn't know if this would add extra bloat for other use cases.

enelson1001:
If I remove the nullptr test and alignment test from DmaFixedBuffer and create a SpiDmaFixedBuffer class like the following

``` C
#pragma once

#include <smooth/core/logging/log.h>

using namespace smooth::core::logging;

namespace smooth::core::io::spi
{
    template<typename T, size_t Size>
    class SpiDmaFixedBuffer
    {
        public:
            static constexpr const char* TAG = "SpiDmaFixedBuffer";
            
            constexpr size_t size()
            {
                return sizeof(T) * Size;
            }


            SpiDmaFixedBuffer()
            {
                // The SPI Master requires the DMA data to be length of multiples of 32 bits to avoid heap corruption.
                static_assert(Size % 4 == 0, "spi dma buffer must have a length of multiple of 32 bits, i.e. 4 bytes");

                smooth::core::util::DmaFixedBuffer<uint8_t, 16> buff(MALLOC_CAP_DMA);
            }

            ~SpiDmaFixedBuffer() = default;
            

            T* data()
            {
                return buff;
            }

            const T& operator[](size_t index) const
            {
                return buff[index];
            }

            T& operator[](size_t index)
            {
                return buff[index];
            }

            bool is_buffer_allocated()
            {
                bool res = buff != nullptr;

                if (!res)
                {
                    Log::warning(TAG, "spi dma buffer has NOT been allocated");
                }
                return res;
            }

            void is_buffer_address_32_bit_aligned()
            {
                // Provide user a warning about address alignment of the buffer
                if (reinterpret_cast<uint32_t>(&buff) % 4 != 0)
                {
                    Log::warning(TAG, "spi dma buffer is not 32 bit alligned, spi transaction efficiency will be reduced");
                }
            }

            bool is_valid()
            {
                is_buffer_address_32_bit_aligned();
                return is_buffer_allocated();
            }


        private:
            T* buff;
    };
}
```

And then in BME280SPI do the following.

``` C
SpiDmaFixedBuffer<uint8_t, 16> rxdata;
bool res = rxdata.is_valid();

if(res)
{
    do stuff with rxdata
}
```

What do you think

PerMalmberg:
I think that's a bad idea. First, you can do 
`using SpiDMAFixedBuffer = DMAFixedBuffer<uint8_t, 4>;` to make a new type without having to implement the whole thing again. Also, you do your should be able to create the buffers you need at start-up of your application instead of as you need them. If you do that then you can instead halt the application if memory allocation fails. This will let you know you have a memory problem during development instead of when you've deployed the device into the field.

enelson1001:
What if I did this.
```C
DmaFixedBuffer.h
#pragma once

#include <algorithm>
#include <heap/include/esp_heap_caps.h>
#include <smooth/core/logging/log.h>

namespace smooth::core::util
{
    template<typename T, size_t Size>
    class DmaFixedBuffer
    {
        public:
            constexpr size_t size()
            {
                return sizeof(T) * Size;
            }

            // Constructor
            DmaFixedBuffer(const uint32_t malloc_cap_type)
            {
                buff = static_cast<T*>(heap_caps_malloc(Size, malloc_cap_type));

            }

            // Destructor
            ~DmaFixedBuffer()
            {
                heap_caps_free(buff);
            }

            /// \return Returns pointer to first element in buffer
            T* data()
            {
                return buff;
            }

            const T& operator[](size_t index) const
            {
                // Prevent going outside buffer
                return buff[std::max(0u, std::min(size() - 1, index))];
            }

            T& operator[](size_t index)
            {
                // Prevent going outside buffer
                return buff[std::max(static_cast<size_t>(0), std::min(size() - 1, index))];
            }

        protected:
            T* buff;
    };
}
```
And derived SpiDmaFixedBuffer from DmaFixedBuffer

``` C
SpiDmaFixedBuffer.h
#pragma once

#include <smooth/core/logging/log.h>

using namespace smooth::core::logging;

namespace smooth::core::io::spi
{
    template<typename T, size_t Size>
    class SpiDmaFixedBuffer : public smooth::core::util::DmaFixedBuffer<T, Size>
    {
        public:
            static constexpr const char* TAG = "SpiDmaFixedBuffer";

            // Constructor derived class of DmaFixedBuffer
            SpiDmaFixedBuffer() : smooth::core::util::DmaFixedBuffer<T, Size>(MALLOC_CAP_DMA)
            {
                // The SPI Master requires the DMA data to be length of multiples of 32 bits to avoid heap corruption.
                static_assert(Size % 4 == 0, "spi dma buffer must have a length of multiple of 32 bits, i.e. 4 bytes");
            }

            // Destructor
            ~SpiDmaFixedBuffer() = default;

            /// Checks to see if buffer has been allocated
            /// \return Returns true if buffer is allocated; false if buffer is not allocted (no space on heap)
            bool is_buffer_allocated()
            {
                bool res = this->data() != nullptr;

                if (!res)
                {
                    Log::warning(TAG, "spi dma buffer has NOT been allocated");
                }

                return res;
            }

            /// Checks if buffer address is aligned on a 32 bit boundary
            void is_buffer_address_32_bit_aligned()
            {
                if (reinterpret_cast<uint32_t>(&this->buff) % 4 != 0)
                {
                    Log::warning(TAG,
                "spi dma buffer is not 32 bit alligned, spi transaction efficiency will be reduced");
                }
            }

            /// Provide user a method to check if buffer is valid
            /// \return Retuns true if buffer is valid false if something went wrong
            bool is_valid()
            {
                is_buffer_address_32_bit_aligned();
                return is_buffer_allocated();
            }
    };
}
```
and then in user program do the following which will catch size error at compile time.
``` C
SpiDmaFixedBuffer<uint8_t, 16> rxdata;
bool res = rxdata.is_valid();

if(res)
{
    Do stuff with SpiDmaFixedBuffer
}
```

PerMalmberg:
> and then in user program do the following which will catch size error at compile time.

No it won't, that check will be done at run-time. Also, is_valid doesn't check the size of the buffer as it is currently written.

lib/smooth/application/display/ILI9341.cpp
```diff
@@ -0,0 +1,399 @@
...
+        core::util::DmaFixedBuffer<uint8_t, 16> rxdata;
+
+        bool res = false;
+
+        if (param_count == 1)
+        {
+            res = read(cmd, rxdata.data(), 1);
+            data.push_back(rxdata[0]);
+        }
+        else if ((param_count > 1) & (param_count < 16))
```
PerMalmberg: No need for extra (), and I think you meant to type &&.

lib/smooth/application/display/ILI9341.cpp
```diff
@@ -0,0 +1,399 @@
...
+            trans.at(0).tx_data[0] = cmd;
+            trans.at(0).flags = SPI_TRANS_USE_TXDATA;
+
+            // setup pre and post control pin states
+            dc_pretrans_pin_states.at(0) = PIN_LOW;
+            cs_pretrans_pin_states.at(0) = PIN_LOW;
+            cs_posttrans_pin_states.at(0) = PIN_LOW;  // keep chip select low
+
+            // configure read parameters transaction
+            trans.at(1).rx_buffer = const_cast<uint8_t*>(rxdata);
```
PerMalmberg: Avoid cont_cast at all cost. The function signature says it is const uint8_t, but here you say it is not cost, which means you open it up for modification. So will it be modified or not? Who knows? Make the function take a non-const parameter instead or find another solution.

PerMalmberg:
I had a look at how you use the fixed buffers. From what I can see you can easily place two or three as members of the class and not having to worry about it any more. Also, remember that the way you currently use them results in memory allocation/deallocations whenever you use the buffers in the respective methods.

enelson1001:
I hope you don't mind a little further discussion on the following.
*You said -> No it won't, that check will be done at run-time. Also, is_valid doesn't check the size of the buffer as it is currently written.*

I have tested my programs using SpiDmaFixedBuffer and DmaFixedBuffer and using SpiDmaFixedBuffer does catch the array being the wrong size at compile time.  As I understand it we can't catch the nullptr at compile time and we can't catch buffer address is not 32 bit aligned at compile time.  If I am wrong I deeply deeply apologize.
The is_valid() function does what I think we would like and that is print a warning message saying the buffer address is not 32 bit aligned and the return true if buffer has been allocated or false if nullptr.

I think that having DmaFixedBuffer generic without any tests inside it is a good approach.  I am not sure all DMA  memory needs have a minimum size of 32 bits or the address of the first item(element) needs to have its address on a 32 bit boundary.  I would think the user program would have to test if dmaFixedBuffer was allocated (nullptr test).

I tested my lvgl + smooth demo program and it works with SpiDmaFixedBuffer and DmaFixedBuffer as shown below.

I have made some minor changes so I have included the files again.

```C
SpiDmaFixedBuffer.h
#pragma once

#include <smooth/core/logging/log.h>

using namespace smooth::core::logging;

namespace smooth::core::io::spi
{
    template<typename T, size_t Size>
    class SpiDmaFixedBuffer : public smooth::core::util::DmaFixedBuffer<T, Size>
    {
        public:
            static constexpr const char* TAG = "SpiDmaFixedBuffer";

            // Constructor derived class of DmaFixedBuffer
            SpiDmaFixedBuffer() : smooth::core::util::DmaFixedBuffer<T, Size>(MALLOC_CAP_DMA)
            {
                // The SPI Master requires the DMA data to be length of multiples of 32 bits to avoid heap corruption.
                static_assert(Size % 4 == 0, "spi dma buffer must have a length of multiple of 32 bits, i.e. 4 bytes");
            }

            // Destructor
            ~SpiDmaFixedBuffer() = default;

            /// Checks to see if buffer has been allocated
            /// \return Returns true if buffer is allocated; false if buffer is not allocted (no space on heap)
            bool is_buffer_allocated()
            {
                bool res = this->data() != nullptr;

                if (!res)
                {
                    Log::error(TAG, "spi dma buffer has NOT been allocated");
                }

                return res;
            }

            /// Checks if buffer address is aligned on a 32 bit boundary
            void is_buffer_address_32_bit_aligned()
            {
                if (reinterpret_cast<uint32_t>(&this->buff) % 4 != 0)
                {
                    Log::warning(TAG,
                "spi dma buffer is not 32 bit alligned, spi transaction efficiency will be reduced");
                }
            }

            /// Provide user a method to check if buffer is valid
            /// \return Retuns true if buffer is valid false if something went wrong
            bool is_valid()
            {
                is_buffer_address_32_bit_aligned();
                return is_buffer_allocated();
            }
    };
}
```

Here is the DmaFixedBuffer.h
```C
DmaFixedBuffer.h
#pragma once

#include <algorithm>
#include <heap/include/esp_heap_caps.h>
#include <smooth/core/logging/log.h>

namespace smooth::core::util
{
    template<typename T, size_t Size>
    class DmaFixedBuffer
    {
        public:
            constexpr size_t size()
            {
                return sizeof(T) * Size;
            }

            // Constructor
            DmaFixedBuffer(const uint32_t malloc_cap_type)
            {
                buff = static_cast<T*>(heap_caps_malloc(Size, malloc_cap_type));
            }

            // Destructor
            ~DmaFixedBuffer()
            {
                heap_caps_free(buff);
            }

            /// \return Returns pointer to first element in buffer
            T* data()
            {
                return buff;
            }

            const T& operator[](size_t index) const
            {
                // Prevent going outside buffer
                return buff[std::max(0u, std::min(size() - 1, index))];
            }

            T& operator[](size_t index)
            {
                // Prevent going outside buffer
                return buff[std::max(static_cast<size_t>(0), std::min(size() - 1, index))];
            }

        protected:
            T* buff;
    };
}
```
Portion of my user program using lvgl and ili9341
```C
static SpiDmaFixedBuffer<uint8_t, MAX_DMA_LEN> display_buf1;
static SpiDmaFixedBuffer<uint8_t, MAX_DMA_LEN> display_buf2;

if (display_buf1.is_valid() && display_buf2.is_valid() )
{
        static lv_color1_t* buf1 = reinterpret_cast<lv_color1_t*>(display_buf1.data());
        static lv_color1_t* buf2 = reinterpret_cast<lv_color1_t*>(display_buf2.data());

        // initialize a descriptor for the buffer
        static lv_disp_buf_t disp_buf;

        lv_disp_buf_init(&disp_buf, buf1, buf2, MAX_DMA_LEN/COLOR_SIZE);
                
        // initialize and register a display driver
        lv_disp_drv_t disp_drv;
        lv_disp_drv_init(&disp_drv);
        disp_drv.buffer = &disp_buf;
        disp_drv.flush_cb = ili9341_flush_cb;  
        lv_disp_drv_register(&disp_drv);

        // LittlevGL graphics library's tick - runs every 5ms
        esp_register_freertos_tick_hook(lv_tick_task);
}
else
{
        ili9341_initialized = false;
}

**Comment by PerMalmberg** [comment]
> I hope you don't mind a little further discussion on the following.

Not at all.

>  As I understand it we can't catch the nullptr at compile time and we can't catch buffer address is not 32 bit aligned at compile time.

That is correct, which makes me think I read too much into what you wrote; namely that that is_valid() would do compile-time checking, which it won't.

> I would think the user program would have to test if dmaFixedBuffer was allocated (nullptr test).

Not necessarily.

All your buffer instances are currently created on the stack (plus he dynamically allocated heap memory), **always with the same size**. This fact allows us to optimize the use of the DMA buffer such that:

- They are created at start-up of the application, or at least when the SPI driver is instanced which likely happens early in the start-up sequence.
- They do the size checks at construction, and if it fails the program is aborted.
- They do the alignment check at construction.

Aborting the application might seem like a bad idea at first, but look at it this way: If there isn't enough (DMA) memory available when you need it you have exhausted the memory and cannot perform the operation you wanted. You're now in a situation you likely cannot recover from; there might not even be enough memory available to log the error.

By having the buffers perform the checks during construction you achieve the following:

- As soon as you have exhausted the memory the application will start aborting. It doesn't happen during compile time, but it does happens as soon as it is possible, i.e. at start-up.
- There's one less check to be done by the application programmer, if the application runs, s/he knows that the requested memory is available.

This is the philosophy I've tried to keep to when designing Smooth to avoid unpleasant surprises once the device is running.

**Comment by enelson1001** [comment]
Thank you for the explanation but couldn't tell which approach you
preferred.
So what would you like me to do?
1. Should I keep the current implementation of the DMAFixedBuffer which
includes the nullptr test and the test for the 32 bit aligned address?
2. Or should I implement the SpiDmaFixedBuffer + DmaFixedBuffer?

I will be happy to implement whatever you choose.

**Comment by PerMalmberg** [comment]
Option two, where the checks are performed as early as possible.

Btw, re. the 32-bit alignment; how do Espressif handle that? We should be able to do the same.

**Comment by enelson1001** [comment]
*Btw, re. the 32-bit alignment; how do Espressif handle that? We should be able to do the same.*

1. It looks like to me they try another malloc????   
2. So should we just let Espressif handle it or what would you suggest.

``` C
esp32-idf   spi_master.c lines 845 to 884
static SPI_MASTER_ISR_ATTR esp_err_t setup_priv_desc(spi_transaction_t *trans_desc, spi_trans_priv_t* new_desc, bool isdma)
{
    *new_desc = (spi_trans_priv_t) { .trans = trans_desc, };

    // rx memory assign
    uint32_t* rcv_ptr;
    if ( trans_desc->flags & SPI_TRANS_USE_RXDATA ) {
        rcv_ptr = (uint32_t *)&trans_desc->rx_data[0];
    } else {
        //if not use RXDATA neither rx_buffer, buffer_to_rcv assigned to NULL
        rcv_ptr = trans_desc->rx_buffer;
    }
    if (rcv_ptr && isdma && (!esp_ptr_dma_capable(rcv_ptr) || ((int)rcv_ptr % 4 != 0))) {
        //if rxbuf in the desc not DMA-capable, malloc a new one. The rx buffer need to be length of multiples of 32 bits to avoid heap corruption.
        ESP_LOGI( SPI_TAG, "Allocate RX buffer for DMA" );
        rcv_ptr = heap_caps_malloc((trans_desc->rxlength + 31) / 8, MALLOC_CAP_DMA);
        if (rcv_ptr == NULL) goto clean_up;
    }
    new_desc->buffer_to_rcv = rcv_ptr;

    // tx memory assign
    const uint32_t *send_ptr;
    if ( trans_desc->flags & SPI_TRANS_USE_TXDATA ) {
        send_ptr = (uint32_t *)&trans_desc->tx_data[0];
    } else {
        //if not use TXDATA neither tx_buffer, tx data assigned to NULL
        send_ptr = trans_desc->tx_buffer ;
    }
    if (send_ptr && isdma && !esp_ptr_dma_capable( send_ptr )) {
        //if txbuf in the desc not DMA-capable, malloc a new one
        ESP_LOGD( SPI_TAG, "Allocate TX buffer for DMA" );
        uint32_t *temp = heap_caps_malloc((trans_desc->length + 7) / 8, MALLOC_CAP_DMA);
        if (temp == NULL) goto clean_up;

        memcpy( temp, send_ptr, (trans_desc->length + 7) / 8 );
        send_ptr = temp;
    }
    new_desc->buffer_to_send = send_ptr;

    return ESP_OK;
```

``` C
esp_ptr_dma_capable located in soc_memory_layout.h lines 143 to 146
inline static bool IRAM_ATTR esp_ptr_dma_capable(const void *p)
{
    return (intptr_t)p >= SOC_DMA_LOW && (intptr_t)p < SOC_DMA_HIGH;
}
```

``` C
defines located in soc.h lines 274 to 276
// Region of memory accessible via DMA. See esp_ptr_dma_capable().
#define SOC_DMA_LOW  0x3FFAE000
#define SOC_DMA_HIGH 0x40000000
```

PerMalmberg:
I see only checks if the buffers is DMA capable, nothing about 32bit
alignment? I'm just about to go to bed so I might be missing it...

enelson1001:
```C
uint32_t* rcv_ptr;
((int)rcv_ptr % 4 != 0))
```
**I thought the above code was testing at the address of the first item in the buffer to be on 32 bit boundary?**

*Other questions you have asked in past emails*
``` C
In lib/smooth/core/io/spi/SPIDevice.cpp:

> +        transaction.user = this;
+
+        auto res = spi_device_polling_transmit(dev, &transaction);
+
+        if (res != ESP_OK)
+        {
+            Log::error(log_tag, "polling_write() failed");
+        }
+
+        return res == ESP_OK;
+    }
+
+    bool SPIDevice::queue_transaction(spi_transaction_t& transaction)
+    {
+        // Attach ourselves as the user-data
+        transaction.user = this;
```
Are there any threading issues doing this? What if a SPI device is destroyed while a transmission is in progress?

**Answer: This is beyond my knowledge base.  I have no idea, sorry**

``` C
In lib/smooth/include/smooth/application/io/spi/BME280Core.h:

> +                                                                 "Forced",
+                                                                 "Forced",
```
Two "Forced"?

**Answer: In datasheet page 15, sensor mode Forced can be either 0x01 or 0x02.  So I thought it would be better have a string for that 0x02 value just in case**

PerMalmberg:
> ((int)rcv_ptr % 4 != 0))

Ah yes, there it is. Notice how they do just a single allocation - does that mean that using MALLOC_CAP_DMA guarantees it being on a 32-bit boundary? 

> This is beyond my knowledge base. I have no idea, sorry

Since the transaction is done in hardware, there is a potential issue here. I think the destructor needs to ensure that any ongoing transaction has completed.

> have a string for that 0x02 value just in case

Ah, ok. It just looked like a copy-paste error.

lib/smooth/include/smooth/core/util/DmaFixedBuffer.h
```diff
@@ -31,32 +31,19 @@ namespace smooth::core::util
...
-                // of the program will suffer since extra coping will be required.
-                if (reinterpret_cast<int>(&buff) % 4 != 0)
-                {
-                    smooth::core::logging::Log::warning("DmaFixedBuffer:",
-                                                        "Not 32 bit alligned, spi transaction efficiency will be reduced");
-                }
             }
 
+            // Destructor
             ~DmaFixedBuffer()
```
PerMalmberg: This destructor must be marked virtual, or the class should be remade such that it cannot be created on the heap (i.e. deleted via pointer).

See "virtual destructors" https://en.cppreference.com/w/cpp/language/destructor

lib/smooth/application/io/spi/BME280Core.cpp
```diff
@@ -0,0 +1,214 @@
...
+{
+    static const char TAG[] = "BME280Core:";
+
+    // Constructor
+    BME280Core::BME280Core()
+    {
+    }
+
+    // Get config_sensor datagram that contains register addresses and data that will be used to configure the BME280
+    void BME280Core::get_configure_sensor_datagram(std::vector<uint8_t>& datagram,
```
PerMalmberg: This might be inherited from my old code, but lets improve it. Using a vector to pass six uint8_t seems overkill. Lets make `datagram` into a `std::array<uint8_t, 6>`

lib/smooth/include/smooth/core/util/DmaFixedBuffer.h
enelson1001: OK completed

lib/smooth/application/io/spi/BME280Core.cpp
enelson1001: The I2CMasterDevices expects a vector for it's writes.
I could modify io/i2c/BME280.cpp use the std::array datagram and copy to a vector that then can be used as the write.
```
io/i2c/BME280.cpp (existing - not modified yet)

bool BME280::configure_sensor(BME280Core::SensorMode mode,
                                  BME280Core::OverSampling humidity,
                                  BME280Core::OverSampling pressure,
                                  BME280Core::OverSampling temperature,
                                  BME280Core::StandbyTimeMS standby,
                                  BME280Core::FilterCoeff filter_coeff,
                                  BME280Core::SpiInterface spi_interface)
    {
        std::vector<uint8_t> datagram;

        bme280_core.get_configure_sensor_datagram(datagram,
                                                  mode,
                                                  humidity,
                                                  pressure,
                                                  temperature,
                                                  standby,
                                                  filter_coeff,
                                                  spi_interface);

        return write(address, datagram, true);
    }
```
Is that what you prefer or leave implenation in BME280Core as vector?

PerMalmberg: Oh, right. Silly me.

No, let it be as-is.

PerMalmberg:
> I will need some time to research this doesn't look that simple to me.

I think all you have to do is call `wait_for_transaction_to_finish()` in the destructor. It'll timeout if there are no transactions ongoing. Sure, waiting 1s for the destruction to complete is a long time, but how often do you destroy an SPIDevice instance?

enelson1001:
Got a response from ESP forum and have updated SpiDmaFixedBuffer.

**Response from ESP forum**
*All results of heap_caps_malloc() will be 32-bit aligned pointers. This is a C standard requirement (results of malloc must be suitably aligned for any variable which can be held in it, and we have 32-bit word size).*

*In the future the implementation could conceivably change so that malloc()-ing less than 4 bytes results in unaligned result (possible but unlikely), however as long as the size you're allocating is a multiple of 4 then the C standard should guarantee the result is a 32-bit aligned pointer.*

Hopefully we are pretty close now.
1. Should I change the name of BME280SPI to BME280SPI4W to make it clear that this class is for 4 line/ 4 wire spi and not 3 line/ 3 wire?  Implementation for 3line is completely different ie no DMA, can only use the transaction data slots ie trans.rx_data[0] ---- trans.tx_data[0].

PerMalmberg:
Ah, learned something new today :)

Nah, a comment in the source clearly stating the fact will do nicely if it is not already there.

enelson1001:
I don't know what has to be done in "This branch has conflicts that must be resolved" .  I fixed the includes not sure what else needs to be done.

PerMalmberg:
Either press the "resolve conflict" button and use the web-editor, or merge master into your branch and reserve the conflicts locally then push to GH.

Instead of changing the includes manually like you did you could have merged master into you branch and it'd have done it for you.

enelson1001:
1. Pull request check - Failed on Master.h saying there is no driver/spi_common.h but there is in esp-idf?
2. Formatting check - Used your uncrustify.cfg to format files but fails on file size changed?

PerMalmberg:
1 - It's when building for Linux, mock-idf doesn't have that file.

2 - Might be different versions of uncrustify, I'll have a look.

PerMalmberg:
I've updated the build system and master to use uncrustify-0.70.0 which is the latest release. I've also merged master into this PR, so you'll have to do a `git pull` to update you local copy. You can then fix the formatting on these files using the new uncrustify config and the named version.

/src/smooth/include/smooth/core/io/spi/SpiDmaFixedBuffer.h
/src/smooth/include/smooth/application/io/spi/BME280Core.h
/src/smooth/include/smooth/application/io/i2c/BME280.h

I've not added the missing file in mock-idf, let me know if you have problems doing that.

enelson1001:
1. It looks like the merge failed? Has a red x next to it.  Maybe because I changed io/i2c/BME280.cpp from your original (changed to use BME280Core.cpp).
2. For mock-idf it looks like I just need to put spi_common.h to the include/driver folder. I will do once we get the merge fixed (if it can be fixed).

PerMalmberg:
No, the merge is fine, it's the same issues as with your commit - formatting and missing file.

enelson1001:
Not sure why it can't find the file? error: esp32/rom/lldesc.h: No such file or directory

enelson1001:
Not sure why mock-idf can't find the file? error: esp32/rom/lldesc.h: No such file or directory

PerMalmberg:
I think you've misunderstood the purpose of mock-idf. Instead of copying the entire file from esp-idf, only add (mostly)empty implementations of the functions actually being used, with return values that won't cause an application calling them to crash.

The idea is that you should be able to compile your application on Linux and debug parts that are not depending on hardware etc. So I think you can remove 90% of the file as well as the lldesc.h file.

PerMalmberg:
You know that you can run the CI-chain locally, right? No need to push to GitHub for each try.

From the root-folder of Smooth: `./CI/build_test.sh` 

Requires a Linux environment with docker installed.

lib/smooth/application/io/i2c/BME280.cpp
enelson1001: Updated

lib/smooth/application/io/spi/BME280SPI.cpp
enelson1001: Updated BME280SPI.cpp

lib/smooth/application/io/spi/BME280SPI.cpp
enelson1001: Updated

lib/smooth/core/io/spi/SPIDevice.cpp
enelson1001: Updated SPIDevice.cpp

lib/smooth/include/smooth/application/display/ILI9341.h
enelson1001: Updated ILI9341.h

lib/smooth/include/smooth/application/io/spi/BME280Core.h
enelson1001: Explained why two forced were used, referenced datasheet

lib/smooth/include/smooth/application/io/spi/BME280Core.h
enelson1001: Done

lib/smooth/application/io/spi/BME280SPI.cpp
enelson1001: Went with different appoach

PerMalmberg:
There's one last small thing - add yourself to the contributors.md file :)

I'll be running a few tests on the I2C parts for the BME280 before merging, unless you already have done that?

enelson1001:
Added my name to contributors but assumed you add the PR # and the rest.

PR #127, opened 2020-04-07
Modified MIMEParser for uploading large files and multiple files

squonk11:
With this PR I would like to make a proposal for modifications on the MIMEParser Class. 
Here I just added mainly the MIMEParser::myparse() function and left the original MIMEParser::parse() in place in order to leave the original code as much as possible untouched. I also added a new test called http_server_test2 which demonstrates the new functionality.
Finally we need to agree how to make the interface of this class. Then the new test probably can be deleted.


lib/smooth/application/network/http/regular/MIMEParser.cpp
```diff
@@ -112,6 +150,176 @@ namespace smooth::application::network::http::regular
...
+                            // std::advance (e, std::distance<std::vector<uint8_t>::const_iterator>(e, b ) );
+                            // Log::info("end of file",".");
+                            // log_data(data.begin(), e);
+                            // Log::info("the end",".");
+                            status = Status::Headers;
+                            data.erase(data.begin(), b - 2);
+                            first_part = false;
+
+                            if (std::distance<std::vector<uint8_t>::const_iterator>(data.begin(),
+                            find_end_boundary()) < 5)
```
PerMalmberg, line 252: What's this magic number 5?

lib/smooth/application/network/http/regular/MIMEParser.cpp
```diff
@@ -112,6 +150,176 @@ namespace smooth::application::network::http::regular
...
+                            content_callback(id, filename, data.cbegin(), b - 2, first_part, true);  // first_b-2 to
+                                                                                                     // avoid additional
+                                                                                                     // crlf in file end
+                            // std::vector<uint8_t>::iterator e;
+                            // std::advance (e, std::distance<std::vector<uint8_t>::const_iterator>(e, b ) );
+                            // Log::info("end of file",".");
+                            // log_data(data.begin(), e);
+                            // Log::info("the end",".");
+                            status = Status::Headers;
+                            data.erase(data.begin(), b - 2);
```
PerMalmberg, line 248: I assume the magic two here is the same as the one above?

Since `b -2`is used twice, lets do something like `const auto end = b - 2` and use `end` instead. Also, lets use the constant mentioned above.

lib/smooth/application/network/http/regular/MIMEParser.cpp
```diff
@@ -112,6 +150,176 @@ namespace smooth::application::network::http::regular
...
+            Data,
+            End
+        }
+        status = Status::Begin;
+        static std::string id{};
+        static std::string filename{};
+        BoundaryIterator begin{};
+        BoundaryIterator end{};
+        static bool first_part = false;
+        static bool end_of_transmission = false;
```
PerMalmberg, line 169: These static, can they? If there are multiple instances of the MIMEParser (as will happen if you have multiple http-server on different threads), then they'd overwrite these static variables. They need to be made into class-members instead.

lib/smooth/application/network/http/regular/MIMEParser.cpp
```diff
@@ -112,6 +150,176 @@ namespace smooth::application::network::http::regular
...
+
+                    if (equal_sign == part.cend())
+                    {
+                        std::copy(std::make_move_iterator(part.begin()),
+                                  std::make_move_iterator(part.end()),
+                                  std::back_inserter(data));
+                    }
+                    else
+                    {
+                        auto key_value = split(part, std::vector<uint8_t>{ '=' });
```
PerMalmberg, line 299: I know this is from my original code, but lets improve it to avoid creating new vector instances each time:
New class-member:
```c++
const std::vector<uint8_t> equal{ '=' };
```

Then:
```c++
auto key_value = split(part, equal);
```

lib/smooth/application/network/http/regular/MIMEParser.cpp
```diff
@@ -82,6 +84,42 @@ namespace smooth::application::network::http::regular
         return mode != Mode::None;
     }
 
+    auto MIMEParser::find_boundary() const
+    {
+        BoundaryIterator p = data.cend();
+
+        if (data.size() > bound.size() + 2)
```
PerMalmberg, line 91: Can't I believe I left magic numbers in my code.....

Lets replace `2` with a constant `LEN_OF_CRLF`

lib/smooth/application/network/http/regular/MIMEParser.cpp
```diff
@@ -198,6 +406,14 @@ namespace smooth::application::network::http::regular
         }
     }
 
+    MIMEParser::BoundaryIterator MIMEParser::my_get_end_of_boundary(BoundaryIterator begin)
+    {
+        // Adjust for CRLF at beginning of boundary pattern
+        auto offset = is_crlf(begin) ? 2 : 0;
```
PerMalmberg, line 412: 2 -> LEN_OF_CRLF

test/http_server_test2/http_server_test2.cpp
```diff
@@ -0,0 +1,363 @@
...
+
+            auto form_data = [uploads, &last_part, &response](const std::string& name,
+                                                        const std::string& actual_file_name,
+                                                        const MIMEParser::BoundaryIterator& begin,
+                                                        const MIMEParser::BoundaryIterator& end,
+                                                        const bool file_start,
+                                                        const bool file_close) {
+                // Store the file in web_root/uploads
+                static Path path{};
+                static std::ofstream to_save;
```
PerMalmberg, line 302: These can't be static, I think - same reasoning as in MIMEParser re. multithreading. Since this is a lambda, they have to be passed in by reference so they can be reused over several calls (and must thus be kept alive in the outside scope).

test/http_server_test2/http_server_test2.cpp
```diff
@@ -0,0 +1,363 @@
...
+                    }
+                    // Log::info("form_data", "before open");
+                    to_save.open(path.str(), std::ios::out | std::ios::app | std::ios::binary) ;
+                    // Log::info("form_data", "File opened");
+                }
+
+
+                auto len = std::distance(begin, end);
+                Log::info("form_data", "chunk to write: {},", static_cast<int>(len));
+                to_save.write(static_cast<const uint8_t*>(&*begin), static_cast<int>(len));
```
PerMalmberg, line 323: This is pretty ugly. The BoundaryIterator can provide a `const uint8_t data()` method for cleaner code.

Alternatively, a helper class that takes the two B-iterators and provides the above method + `size_type len() const`

squonk11:
I was very busy the last weeks due to several job and private reasons, so I was not able to work on this topic. Luckily I was/am not ill - but I need to take care. The situation in Sweden currently seems to be less critical than in Germany.
The "memory leak" (which probably was not a memory leak) seems to be fixed now. I moved the line `parse_content_disposition(headers, content_disp);` in function `MIMEParser::consume_headers` out of the `while (std::getline(ss, s, '\r'))` loop. Since then it seem so to work - I did some testing already and did not see the crash any more. Although, I don't 100% understand what was going wrong there. I did this step because I saw that the crash happened in `parse_content_disposition()` and I realized that it is being called for each cycle of the while-loop - which is not necessary.
Thank you for providing the various hints and comments - very appreciated. I will go through them hopefully during the next days.
Is there a way to do the CI-build locally on my machine? Making code changes, pushing them to the PR and waiting for the result is quite cumbersome. Making this locally would speed up the job.

PerMalmberg:
Wait there's a crash and a possible memory leak? Not having to call the function each turn of the loop is good, but that really only reduces the chance of it happening since less code is run.

Yes, you can run the CI steps locally. The commands to run, from the project root, can be seen here:
https://github.com/PerMalmberg/Smooth/blob/master/.github/workflows/ci.yml

squonk11:
Yes, I agree that more testing is needed. I will try to make a test several times with maybe 10000 files. In order to reduce the runtime of this, I could simply not write the data to the SD-card as this also is a bottleneck.

Thank you for the hint on running the CI step locally - but probably this is possible for Linux only, I guess. I am on Windows. I might try to use the Windows Subsystem for Linux - but I am afraid this might take several hours of work for installing all required tools and realizing in the end that I don't get it running.

PerMalmberg:
The CI process uses Docker so I believe you can run the CI-scripts from WSL if you have Docker installed in WSL (I don't know if that is possible)

For the formatting, uncrustify runs on Windows too, just make sure to use the same version (0.70.0) as used by CI and take the config from the repo.

test/http_server_test2/http_server_test2.cpp
squonk11: I am not sure if these can't remain static. If I would launch another http-server, there would be other lambda functions defined for managing the uploads, since this is not a function inside the smooth library but a 'potential' user code.
If I would have to pass these in by reference, I think it would be a messy thing passing it from `App` to the lambda function `upload` and from there to the lambda function `form_data`. 
Or am I missing something here?

test/http_server_test2/http_server_test2.cpp
squonk11: to be honest, I don't get it. If I write: 
`to_save.write((char*)&*begin, static_cast<int>(len));`
the code compiles without errors/warning on my machine; and the code also works. Using data() or reinterpret_cast cause various error messages.
Do you have a more specific hint on how to solve it?

lib/smooth/application/network/http/regular/MIMEParser.cpp
squonk11: o.k. fixed

lib/smooth/application/network/http/regular/MIMEParser.cpp
squonk11: o.k. fixed

lib/smooth/application/network/http/regular/MIMEParser.cpp
squonk11: o.k. fixed

lib/smooth/application/network/http/regular/MIMEParser.cpp
squonk11: o.k. fixed

lib/smooth/application/network/http/regular/MIMEParser.cpp
squonk11: o.k. fixed

lib/smooth/application/network/http/regular/MIMEParser.cpp
squonk11: o.k. fixed

squonk11:
Now I made a new test with uploading 8000 files. When I run this test with actually writing these files to the SD-card the test runs approx. 2,5h. If I run it without writing to the SD-card the the complete upload takes approx 1min. So, I think there is also something not o.k with the SD-card access.

test/http_server_test2/http_server_test2.cpp
PerMalmberg: Have a look at http://cpp.sh/9j66l and run it. You'll see that it prints "1", "2" even though the example is making copies of the lambda into `b` and `c`. So, if you have statics inside your lambda it'll be shared between copies, just like a class.

The example reuses the lambdas for both the secure and insecure server, contrary to you point that there will be different lambdas so the lambdas must be thread-safe.

I agree that having to pass additional arguments would be a bit messy. Would it perhaps be better to add a template argument to the HTTPServer class that specifies a type of which an object can be retrieved in the lambda...something like a session object?

test/http_server_test2/http_server_test2.cpp
PerMalmberg: I meant `const uint8_t* data()`, but I suppose you figured that out already.

I'm guessing the problem is different compiler versions or settings. I don't have things setup so I can unfortunately not dig deeper into it at this time. The docker image is from Sept 26, 2019 so whatever xtensa-gcc8 was awailable then is what is being used now. IIRC, it was the second version.. "r2" something? Perhaps it is time to update the docker image.

PerMalmberg:
Opening an closing files is a fairly slow process. 8000 times over 2.5h means it takes 1.125 seconds per file. That is slower than I'd expect. What kind of write speeds to you get with the [SD Test](https://github.com/PerMalmberg/Smooth/blob/master/test/hw_sdcard_test/hw_sdcard_test.cpp)

squonk11:
I will check the result of the SD-Test later. But I think, it is not really a problem with writing to the SD-card - it is a fast SD-card and I activated `host.max_freq_khz = SDMMC_FREQ_HIGHSPEED;` from the beginning.
I did the upload test with 8000 files two more times with writing to the SD card and I had a duration of 1h and 1,5h. So, the upload seems to vary quite a lot. I need to do more testing - but now I have to care about some gardening first.
And until now each time the upload was successful - no system crash and all files correctly written.
An issue which remains is that the success message in the browser after uploading all files does not appear if I upload many files - when uploading less files I see the final message sent by the webserver. Maybe this is related to the fact, that the upload duration is too long and the browser does not wait that long. When uploading 8000 files without writing to the SD card, the success message appears - so, it really seems to be related to the duration.

test/http_server_test2/http_server_test2.cpp
squonk11: yes, you are right. I need to think again about how to solve it.

squonk11: An easy solution could be to pass a void* to the HTTPServer.on() method. But I guess this is more C-style rather than C++, right?

PerMalmberg:
Wifi has a part in the transfer rate too.

squonk11:
1. I made a mistake in my previous test: I forgot to comment in a again the writing to the file and closing it. If I do this correctly, the upload takes 2,5h again. In addition to that I get regularly a watchdog error message during the upload. Probably the watchdog needs to be serviced somewhere.

2. I think I might have problems with solving the issue with the static declaration in the lambda function. I am not feeling strong enough to fix it using the template approach. Is there any other way to solve it?

PerMalmberg:
Well, there probably is, and the template-way was just an idea. Apart from passing the variables as arguments I don't currently have an idea though. Need to think on it a bit.

PerMalmberg:
I've begun setting up my environment, hope to have time to take a look at this tomorrow.

squonk11:
o.k. thanks a lot for taking care of this. 
When you setup the environment based on the the latest esp-idf master or even v4.1 you will probably see that there are compile time warnings:

1. due to `tcpip_adapter_init()` being deprecated which should be replaced by `esp_netif_init()`

2. due to some new(?) Xtensa library functions (e.g. in xtruntime.h) which neglect sign conversion mismatches.

The first topic I could try to care about as soon as the MIMEParser topic is closed. For the 2nd I currently only see the chance to remove the `-Wsign-conversion` directive from compiler_options.cmake.

PerMalmberg:
Yeah, there are quite a few places where the sign-conversion warning results in an error. I've fixed those in 230e2f311ef0c6111f93d9adb962856c835f1259

I'll update to use the new netif-library too, but I went off on a tangent - I'm currently exploring VSCode's new Remote-Container support in which it is possible to run everything inside a docker image, meaning that its easy to share development environments. I think I'm nearing completion of this, then I'll get back to fixing the rest.

PerMalmberg:
Master-branch now uses the latest IDF master as of today. I've also fixed the needed netif stuff and event system changes. Sadly I didn't have time to also do the template-thing today so I'll have to do that during the week.

PerMalmberg:
@squonk11 I just had a look at what the best way to enable the HTTP Server provide a shared a user-defined object between each callback call. I'm currently leaning towards dropping the use of lambdas and instead use objects provided by the user of the HTTPServer. This way the objects can hold whatever they need. I think it will also make the use of the HTTPServer simpler.

Something like a base class like this:

```c++
class RequestHandler
{
public:
    virtual void request(IServerResponse& response,
            IConnectionTimeoutModifier& timeout_modifier,
            const std::string& url,
            bool first_part,
            bool last_part,
            const std::unordered_map<std::string, std::string>& headers,
            const std::unordered_map<std::string, std::string>& request_parameters,
            const std::vector<uint8_t>& content,
            MIMEParser& mime) = 0;

    // Convenience methods that are called before the first and after the last part.
    // Ca be used to reset things etc.
    virtual void called_before_first_part() {}
    virtual void called_after_last_part() {}
};
```

The user can then do this:

```c++
class MyRequestHandler : public RequestHandler
{
public:
    void request(....) override
    {
         if (first)
        {
            hold_my_data["filename"] = .....;
        }
    }
private:
    std::unordered_map<string, string> hold_my_data{};
};
```

It's ofc possible to use any container or other type instead of a map; anything that fits the use-case.

test/http_server_test/http_server_test.cpp
```diff
@@ -40,6 +40,15 @@ using namespace smooth::application::network::http::responses;
...
 {
+
+    extern "C" {
+        void app_main() {
+            App app;
+            app.start();
+        }
+    }
+
+
```
PerMalmberg, line 51: This need to be removed,

test/http_server_test/http_server_test.cpp
```diff
@@ -220,112 +230,131 @@ namespace http_server_test
...
+                        remove(path);
+                    }
+                    Log::info("form_data", "before open");
+                    to_save.open(path.str(), std::ios::out | std::ios::app | std::ios::binary) ;
+                    Log::info("form_data", "File opened");
+                }
+
+                auto len = std::distance(begin, end);
+                Log::info("form_data", "chunk to write: {},", static_cast<int>(len));
+                to_save.write((char*)&*begin, static_cast<int>(len));
```
PerMalmberg, line 325: Had forgotten about you asking about this one. Try this, it compiles on both for Linux and Xtensa.

`to_save.write(reinterpret_cast<const char*>(&*begin), static_cast<int>(len));`

squonk11:
The magic numbers should be eliminated - if not then maybe I lost some of the modifications. I need to check again. 
The suggestion with the RequestHandler looks promising. When you implement this I need to rework my code also - but no problem. Also some tests need to have some rework.

squonk11:
1. concerning magic numbers:
in my first commit yesterday I unintentionally uploaded some old files. The latest push yesterday evening should now include the latest version of the MIMEParser class.

2. upload speed
I did some more testing on the upload speed. It seems that the reason for the slow upload is related to the number of files already in the directory of the SD-card. If I upload into an empty directory, the upload speed for the first 200...300 files is quite fast. If there are already many files available, the upload speed is dramatically reduced. The bottleneck seems to be the opening of the file. So, I guess the problem is related to the handling of the file allocation table in the SD-card driver - maybe the complete file list is loaded into RAM which becomes to be a timing problem when there is few internal RAM left.

PerMalmberg:
What you found re. upload speed is interesting and I think you've concluded what the actual issue is. Good job.

I've implemented the first version of the new request handlers in PR #129. It feels a lot cleaner than before. I'm considering doing something similar for the callback to the MIME-parser, thoughts on that?

squonk11:
I have taken a short look at your changes. IT really looks much more clean and more logical. Doing the same for the MIMEParser probably also makes sense. I can integrate my changes to the MIMEParser afterwards.

PerMalmberg:
I pushed an update yesterday. While I like having a class instead of lambda for the callbacks from MIMEParser, I'm not sure my[ attempt at reducing the number](https://github.com/PerMalmberg/Smooth/pull/129/files#diff-11ec0620cd9039fb56a8fcbe54babb47R68) of arguments in the interface is the best solution. Please have a look and let me know what you think.

squonk11:
I checked the modifications you made online in the Pull Request. It is a bit difficult to read due to fragmented source code shown - probably it would have been easier to pull the complete source and apply the patch of this Pull Request.
But as far as I can see the approach looks o.k. for me. Also the encapsulation of the RequestParms in the HTTPRequestHandler class looks good to me. 
If you check-in these modifications I could apply my modifications to the MIMEParser to it.

PerMalmberg:
It's all merged to master now.

Remember that there's settings for VS Code that lets you run the CI-flow in docker on your windows machine. I've only run it on Linux, but they say it should work equally well on Windows. Instructions here: https://code.visualstudio.com/docs/remote/containers

squonk11:
Thank you for providing the improvements on the invocation of the handlers. 
Unfortunately I am currently again quite busy but I hope I will be able to work on my PR during the next days.

PerMalmberg:
No rush.

You're free to do as you wish ofc, but I'd start a new PR for a fresh start. I find that when so much of the underlying foundation changes it's better to start over, taking into consideration all the things that was learned during the first attempt.

squonk11:
Unfortunately I am not familiar with Docker - I never used it. 
Just for my understanding:

- The Docker host must be a Linux System, right? Do you know if I can use the WSL on my Win10 for it? Or is it possible to use my Win10 system only with Docker Desktop installed (without using WSL)?

- Once I have the Docker host up and running: do I have to setup the all the Espressif toolchain on it or is it already included in some Docker Image I could clone from somewhere (from Espressif or from your repo)?

PerMalmberg:
>     * The Docker host must be a Linux System, right? Do you know if I can use the WSL on my Win10 for it? Or is it possible to use my Win10 system only with Docker Desktop installed (without using WSL)?

No, see the [page I linked earlier](https://code.visualstudio.com/docs/remote/containers).
 
>     * Once I have the Docker host up and running: do I have to setup the all the Espressif toolchain on it or is it already included in some Docker Image I could clone from somewhere (from Espressif or from your repo)?

The entire tool-chain is available in the docker image. You just tell VS Code to open the folder as "remote". Actually, when you have the VSCode extension installed, VSCode asks if it should open it in "remote" mode. The VSCode settings in the repo points to the [prebuilt image I use](https://github.com/PerMalmberg/Smooth/blob/master/.devcontainer/devcontainer.json#L9).

squonk11:
I installed Docker and the plugin for VSCode. But unfortunately I am getting an error message: `[2033 ms] Command failed: docker start 9f04baa6d337c1583e13dbf2d3c9bd523d83f4a9eab2d712dd360cfe2ed8e315` when I try to open the folder. Is there somewhere a log file where I can check was is going wrong?

PerMalmberg:
Docker surely have a log somewhere, but my guess it is this line that causes it:

https://github.com/PerMalmberg/Smooth/blob/master/.devcontainer/devcontainer.json#L30

You'll need to change to whatever port you're using. Lookup the format for a Windows COM-port in the docker docs.

squonk11:
I changed the line to `"runArgs": [ "--device=COM5" ]` but it did not help. Then I removed the docker image using `docker image rm -f vsc-smooth4-face9282dda63c7fbe0e32b232c21b72` hoping that then a new image will be built. But there is still the same error message even though the smooth docker image is not available any more ('docker image ls' does not list a smooth container). Issuing the command 'Rebuild Container' from VSCode seems to do nothing.
Now I started the rebuild of the image via command line. This is currently running. But I have to go to sleep now. I'll see the result tomorrow.

squonk11:
after building the container again I have the same error message. I think I need to read a bit more about Docker.

squonk11:
Now I commented out the line with `runArgs` and rebuild the Docker image again. Then I can start the docker image. It asks me to start the compile process with the command `idf.py build`. If I do so, the compile stops with the error message:

> CMake Error at /esp/esp-idf/tools/cmake/component.cmake:302 (message):
  Include directory '/workspaces/Smooth4/externals/nlohmann_json/include' is
  not a directory.
The directory `./externals/nlohmann_json` is empty.

PerMalmberg:
It asks you?

Dont you mean that it says that you can use idf.py to compile? That what it
does anytime the IDF environment is initialized.

Anyway, I don't know what command you ran, but it looks like your clone of
smooth isn't complete.

Try running

git submodule update --init --checkout --recursive

squonk11:
yes, I noticed already, that the submodules are missing. Now it compiles correctly. But I am not able to map a Windows COM port to the linux docker client. It seems as if this is not possible at all: [https://forums.docker.com/t/how-to-expose-host-serial-port-to-container-correctly/81588/6](https://forums.docker.com/t/how-to-expose-host-serial-port-to-container-correctly/81588/6)
But I think I can work under windows on my local copy as before and just use the docker image to merge the new code in and test before committing the PR.
Now I am asking myself how to compile the `http_server_test`. If I invoke in `./CI` the command `./build_test.sh ../test/http_server_test` I get the error message `bash: ./build_test.sh: /bin/bash^M: bad interpreter: No such file or directory`.

squonk11:
ah, now I see: I need to use `set(selected_test_project starter_example)` in CMakeLists.txt to select the project to compile. Reading educates...

PR #132, opened 2020-05-16
modified MIMEParser class for upload of larger files and multiple files

squonk11:
modified MIMEParser class for uploading lager files and multiple files

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -6,10 +6,10 @@
 	// Sets the run context to one level up instead of the .devcontainer folder.
 	"context": "..",
 
-	// "image": "permalmberg/smooth:latest",
+	"image": "permalmberg/smooth:latest",
 
 	// Update the 'dockerFile' property if you aren't using the standard 'Dockerfile' filename.
-	"dockerFile": "../CI/docker-image/Dockerfile",
+	//"dockerFile": "../CI/docker-image/Dockerfile",
 
 	// Set *default* container specific settings.json values on container create.
 	"settings": { 
@@ -27,7 +27,8 @@
 
 	// Uncomment when using a ptrace-based debugger like C++, Go, and Rust
 	// "runArgs": [ "--cap-add=SYS_PTRACE", "--security-opt", "seccomp=unconfined" ],
-	"runArgs": [ "--device=/dev/ttyUSB1" ]
+	//"runArgs": [ "--device=/dev/ttyUSB1" ]
+	"runArgs": [ "-p", "8080:8080/tcp", "-p", "8443:8443/tcp" ]
 
 	// Uncomment to use the Docker CLI from inside the container. See https://aka.ms/vscode-remote/samples/docker-in-docker.
 	// "mounts": [ "source=/var/run/docker.sock,target=/var/run/docker.sock,type=bind" ],
--- a/.vscode/c_cpp_properties.json
+++ b/.vscode/c_cpp_properties.json
@@ -34,6 +34,7 @@
             "cStandard": "c11",
             "cppStandard": "c++17",
             "defines": [
+                "ESP_PLATFORM=0"
             ],
             "browse": {
                 "path": [
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -1,0 +1,27 @@
+{
+    // Use IntelliSense to learn about possible attributes.
+    // Hover to view descriptions of existing attributes.
+    // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
+    "version": "0.2.0",
+    "configurations": [
+        {
+            "name": "(gdb) Launch",
+            "type": "cppdbg",
+            "request": "launch",
+            "program": "${workspaceFolder}/build/test/http_files_upload_test/http_files_upload_test.elf",
+            "args": [],
+            "stopAtEntry": false,
+            "cwd": "${workspaceFolder}",
+            "environment": [],
+            "externalConsole": false,
+            "MIMode": "gdb",
+            "setupCommands": [
+                {
+                    "description": "Enable pretty-printing for gdb",
+                    "text": "-enable-pretty-printing",
+                    "ignoreFailures": true
+                }
+            ]
+        }
+    ]
+}
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -3,7 +3,7 @@
     "cmake.saveBeforeBuild": true,
     "cmake.parallelJobs": 4,
     "cmake.configureSettings": {
-        "ESP_PLATFORM": 1
+        "ESP_PLATFORM": 0
     },
     "files.associations": {
         "typeinfo": "cpp",
@@ -72,5 +72,6 @@
         "unordered_set": "cpp",
         "future": "cpp",
         "iomanip": "cpp"
-    }
+    },
+    "python.pythonPath": "/usr/local/bin/python"
 }
--- a/CI/docker-image/Dockerfile
+++ b/CI/docker-image/Dockerfile
@@ -25,6 +25,7 @@ RUN apt update \
 		libmbedtls-dev \
 		gcc-8 \
 		g++-8 \
+		gdb \
 		libssl-dev \
 		make \
 		libz-dev \
@@ -50,7 +51,9 @@ RUN wget https://www.python.org/ftp/python/3.8.2/Python-3.8.2.tar.xz \
 	&& cd .. \
 	&& rm -rf Python-3.8.2 \
 	&& rm -rf Python-3.8.2.tar \
-	&& ln -s /usr/local/bin/python3 /usr/local/bin/python
+	&& ln -s /usr/local/bin/python3 /usr/local/bin/python \
+	&& pip3 install --upgrade pip \
+	&& pip3 install pylint requests
 
 # ESP-IDF
 RUN cd / \
@@ -96,5 +99,9 @@ RUN echo ". $IDF_PATH/export.sh" >> ~/.bashrc
 # Set locale so that "idf-py menuconfig" can be run.
 RUN echo 'export LC_ALL="C.UTF-8"'  >> ~/.bashrc && echo 'export LC_CTYPE="C.UTF-8"'  >> ~/.bashrc
 
+EXPOSE 80
+EXPOSE 8080
+EXPOSE 443
+EXPOSE 8443
 
 CMD [ "/bin/bash" ]
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -22,7 +22,7 @@ set(CMAKE_CXX_STANDARD 17)
 
 
 # Select the test project to build
-set(selected_test_project http_files_upload_test)
+set(selected_test_project starter_example)
 
 # For Linux builds, you may enable address sanitizer
 set(SMOOTH_ENABLE_ASAN 0)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -20,8 +20,9 @@ limitations under the License.
 cmake_minimum_required(VERSION 3.10)
 set(CMAKE_CXX_STANDARD 17)
 
+
 # Select the test project to build
-set(selected_test_project starter_example)
+set(selected_test_project http_files_upload_test)
 
 # For Linux builds, you may enable address sanitizer
 set(SMOOTH_ENABLE_ASAN 0)
@@ -39,6 +40,7 @@ list(APPEND available_tests
         server_socket_test
         secure_server_socket_test
         http_server_test
+        http_files_upload_test
         destructing_event_queues
         destructing_subscribing_event_queues
         security
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -5,7 +5,7 @@ Below is the list of individuals that have contributed to the project.
 |Name|Contribution details
 |----|----|
 |[Per Malmberg](https://github.com/PerMalmberg)|Author and maintainer|
-|[squonk11](https://github.com/squonk11)|[PR#65](https://github.com/PerMalmberg/Smooth/pull/65), [PR#114](https://github.com/PerMalmberg/Smooth/pull/114)|
+|[squonk11](https://github.com/squonk11)|[PR#65](https://github.com/PerMalmberg/Smooth/pull/65), [PR#114](https://github.com/PerMalmberg/Smooth/pull/114)|, [PR#132](https://github.com/PerMalmberg/Smooth/pull/132)|
 |[KerryRJ](https://github.com/KerryRJ)|[PR#71](https://github.com/PerMalmberg/Smooth/pull/71)|
 |[jeremyjh](https://github.com/jeremyjh)|[PR#87](https://github.com/PerMalmberg/Smooth/pull/87)|
 |[COM8](https://github.com/COM8)|[PR#98](https://github.com/PerMalmberg/Smooth/pull/98)|
--- a/lib/smooth/application/network/http/regular/MIMEParser.cpp
+++ b/lib/smooth/application/network/http/regular/MIMEParser.cpp
@@ -40,6 +40,7 @@ namespace smooth::application::network::http::regular
         expected_content_length = 0;
         mode = Mode::None;
         end_of_transmission = false;
+        parse_status = ParseStatus::Begin;
     }
 
     bool MIMEParser::detect_mode(const std::string& content_type, std::size_t content_length)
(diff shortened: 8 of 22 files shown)


squonk11:
I will take care about the remaining issues during the next days. Unfortunately I have a problem running the shell scripts for the tests in my Docker; I think this is related to fact that the shell scripts are having CRLF line endings. I will try to find a fix for this also.

squonk11:
Now I changed the line endings of the shell scripts to LF and the error message of the missing interpreter "bin/bash^M" is gone and the script starts to be executed in Docker. But now I get the error message `./CI/build_test.sh: line 14: docker: command not found`. Do you have an idea what I am missing?

PerMalmberg:
The CI-scripts are meant to be run outside of docker; they will start an instance of the docker-image to run the tests.

squonk11:
but starting the scripts outside of Docker I can only do under Linux, right? Or is there a trick to do it under Windows also - maybe WSL?

PerMalmberg:
Yes, I'd expect WSL to work.

I'd guess any bash terminal would work, such as Git-bash that comes with Git for Windows.

squonk11:
unfortunately both do not work. With git bash I get the error message: `C:\Program Files\Docker\Docker\resources\bin\docker.exe: Error response from daemon: the working directory 'C:/Program Files/Git/src' is invalid, it needs to be an absolute path.` I guess this is somehow related to the my Git installation because the directory C:/Program Files/Git/src really does not exist.
With WSL I get the error message `docker: Cannot connect to the Docker daemon at unix:///var/run/docker.sock. Is the docker daemon running?.` This probably is due to the fact that WSL assumes that Docker is installed under WSL (but it is not).
So, I think I have to continue the try and error method via several commit/push sequences.

PerMalmberg:
Aww, sorry to hear that.

You can ofc simply run the commands that the CI-scripts themselves run from within Docker.

PerMalmberg:
Last! :) @squonk11  :)

PerMalmberg:
@squonk11 Are you happy with the PR in its current state? I've not test run it myself, but I presume you have.

squonk11:
I did quite a lot of testing with the previous version using lambda functions. This version I did not test so much. I do not expect issues with this version but before accepting the pull request maybe it would be better to do some more testing. I will try to do it over the next weekend and finally add this PR to the file CONTRIBUTORS.md.
Do you have any more comments on the code?

PerMalmberg:
I didn't see any obvious issues with the code.

I was thinking that we could perhaps do some automatic testing using the Linux version and curl inside docker, i.e. do actual calls to the HTTP-server. That way we get reproducability of the tests and can test a large range of cases.

squonk11:
Automatic testing would be a good idea. Unfortunately I am having problems starting bash scripts from outside Docker. Is it possible to do use these scripts from inside Docker as well? How to start the HTTP-server inside Docker? Additionally I did not use curl before - but I think I could find out how to use it.

PerMalmberg:
The scripts that currently start the CI-test and builds just start a docker instance to run other scripts. The "other" scripts can all be run manually from within docker.

Python might actually be a better idea than curl, as it is easier to build tests in Python than in bash.

Starting the HTTP-server inside docker would require a test application built for testing the HTTP server specifically. The tests we currently have can serve the example folder with all the images, but that's not really useful in automated testing.

I'm thinking that an echo server, i.e. reply back with what is received, could be a useful thing. Then we can automated sending and receiving data from a single byte up to several mega-bytes with little effort.

squonk11:
So, the Python script should launch a HPPT client which sends the files to the HTTP test-server. The HTTP test-server will then echo the files back to the python script. In order to receive the echoed files the python script needs to also launch a HTTP server which will then receive the files and comapres the contents against the files sent. Is that what you have in mind?

PerMalmberg:
No, no need for a HTTP server on the Python side, that can be done over the same client socket that Python uses to send the data. So, POST some data, then receive the answer.

squonk11:
Will the upload from the Python script be done via `requests`; e.g. like so:
```
import requests

url = "http://localhost:5000/"
fin = open('simple_table.pdf', 'rb')
files = {'file': fin}
try:
  r = requests.post(url, files=files)
	print r.text
finally:
	fin.close()
```
?
How must the Http-Server send the file back to the Python script? I guess it must send it also via a Multipart/Mime protocoll. But how to do that - there is not such a functionality in Smooth, isn't it?

PerMalmberg:
Yes, requests will most likely do the job.

No there isn't. How about this:

1. Pre-generate a set of random binary data so that each test can be recreated and re-run.
2. Create a hash of the data
3. Send the data to Smooth.
4. Calculate the hash on the received data.
5. Reply with the hash
6. Compare the hash. If they don't match, the test fails.

This would allow us to test everything from 1 byte to megabytes.

squonk11:
The hash will be one hash value calculated over all files to be send (in case of multiple files)? The hash value will then be returned via the HTTP response?

PerMalmberg:
One per file, and yes it will be the response.

squonk11:
But there is only one response also in the case of multi file upload. So, the reply must contain the hashes for all files; e.g. like this:
"File1.txt:hash1
File2.bin:hash2
File3.xyz:hash3
...
"
Pyhthon then needs to analyze this response. Is that correct?

PerMalmberg:
Oh, right. To keep it as simple as possible I think it's best to return a
JSON object with file/hash pairs for each file. We can then use the same
code regardless if one or hundred files were received/sent.

squonk11:
So, the JSON object will look e.g. like this:
```
([
  {
    "filename": File1.txt,
    "hash": hash1
  },
  {
    "filename": File2.bin,
    "hash": hash2
  },
  ...
)]
``` 
? I guess this can be generated using nlohmann library.

PerMalmberg:
Yes, although there is no need for separate JSON object per file i think.
Just {"x" : "y", "z":"w", ... }

Nlohmann should solve this no problem.

squonk11:
o.k. I can try to setup such a test. But probably I will have questions because until now I have only very low or no experience with Python, JSON, Docker, hash calculation, ... . Additionally I also have not much time to work on this - so, it will take some time. If you can wait for that, then it is o.k. for me.

PerMalmberg:
We're both busy people 🙂

There is no rush getting this done. It's better it does get done and we
know the HTTP implementation is good. This is something I've wanted to do
for a while now for the original implementation but had to put off for
other things.

I'll see what I can do on my end in the near-ish future.

squonk11:
A few more questions before I start to implement the new test function:

1. You mentioned above that this should be a test function using the "linux version". If I use the http_server_test as a basis, shouldn't it work directly for both, linux and ESP32? I have no linux server available, so I can only work on the ESP32.
2. Can I simply implement the non-TLS server just for simplicity?
3. I see there are many algorithms available for calculating hashes. MD5 seems to be used quite frequently. There are libraries for C++ and also for Python (e.g. in hashlib). Is this a good choice in your view, or do you have a better suggestion?
4. Could it make sense to accept the PCR for the modifications on MIMEParser made so far in order to separate them and also our following discussions from another?

PerMalmberg:
>     1. I have no linux server available, so I can only work on the ESP32

That's fine. As long as no hardware, for which there is no mock-up, a Smooth-app can be compiled for both platforms. It's just so much easier and quicker to develop for Linux, and you have the additional benefit of being able to use tools like AdressSanitizer and Valgrind.

> Can I simply implement the non-TLS server just for simplicity?

Thankfully using TLS or not is just a matter of selecting the socket-type for the template argument so the HTTP server is not aware of if it is running over TLS or not. So, yes, but there is really no difference.

> I see there are many algorithms available for calculating hashes...

ESP-IDF comes with libsodium which has support for [blake2b hashes](https://doc.libsodium.org/hashing/generic_hashing) and so [does Python](https://docs.python.org/3/library/hashlib.html). As long as we use the same key on both sides they will produce the same hash. I'd recommend using this, unless you have a better suggestion.

> Could it make sense to accept the PCR for the modifications on MIMEParser made so far in order to separate them and also our following discussions from another?

It does, though I want to at least test-run it myself a bit first.

> I will call the new test function http_files_upload_test. Is that o.k.?

I have no objections to that.

squonk11:
While testing the file upload using a Python script I detected a residual problem that the reply is not being sent because `is_last()` never returns true in `UploadResponder.cpp` because in `HTTPServer.h` last_part is not being passed to the handler for the the last part.
So, don't accept the current status of my PR. I will try to fix this issue first.

PerMalmberg:
Glad to hear the testing is having a positive effect.

squonk11:
I found this problem already recently and I thought I fixed it. But somehow the fix did not find its way into the PR. 
In order to overcome this problem, I simply used an override for `end_of_request()` in order to send the reply. I think, since now there is the `end_of_request()` override available in `HTTPRequestHandler` the previous solution using `is_last()` in order to check if the reply can be sent is obsolete now. Maybe this function can be removed from `HTTPRequestHandler` - also because it does not work since `request_params.last_part` is not updated in `HTTPServer::handle`. BTW the `http_server_test` now in the repo probably also does not work (I did not check).

PerMalmberg:
> In order to overcome this problem, I simply used an override for `end_of_request()` in order to send the reply.

That's a way to do it, but to fix the actual problem I think [this line](https://github.com/PerMalmberg/Smooth/blob/73829f2edcbfa4b362359ae4f6cf93320faab565/lib/smooth/include/smooth/application/network/http/HTTPServer.h#L192) should be copied to *before* [this line](https://github.com/PerMalmberg/Smooth/blob/73829f2edcbfa4b362359ae4f6cf93320faab565/lib/smooth/include/smooth/application/network/http/HTTPServer.h#L203) so that is_last is updated.

squonk11:
> That's a way to do it, but to fix the actual problem I think this line should be copied to before this line so that is_last is updated.

This I tested already and it does not work. The reason is that the update of `request_params.last_part` must be done _before_ the call of `handler->request(timeout_modifier, requested_url, data);` because `is_last()` is called inside `handler->request()`. But doing the `update_call_params()` allways before `handler->request()` could be quite costly(?) because it will be called for each chunk of data received.

PerMalmberg:
Oh, right. So then we do this instead.

```C++
// Call order is important - must update call params before calling the rest of the methods in the
// inheriting class.
if (first_part || last_part)
{
    handler->update_call_params(...)
}

if (first_part)
{
    handler->prepare_mime();
    handler->start_of_request();
}

handler->request(...);

if (last_part)
{
    handler->end_of_request();
}

```

squonk11:
That looks good. I will check it tomorrow.

squonk11:
Now I am trying to calculate the hash values but I am stuck at the very beginning: as soon as I include sodium.h and define `crypto_generichash_state state;` I get the linker error `memalign_function_was_linked_but_unsupported_in_esp_idf`. Obviously sodium requires some feature which esp-idf is not supporting? Do you have an idea?

squonk11:
I encountered another problem: the hash calculated in Python using hashlib is different from value that libsodium on the ESP32 side calculates. Is there any special thing to be taken into account? Currently I am generating a hash  of length 32 (which seems to be the only size which libsodium supports) and I am using the same keystring "my16characterKey".
Do you have any hints?

PerMalmberg:
> I encountered another problem: the hash calculated in Python using hashlib is different from value that libsodium on the ESP32 side calculates. Is there any special thing to be taken into account? Currently I am generating a hash of length 32 (which seems to be the only size which libsodium supports) and I am using the same keystring "my16characterKey".
> Do you have any hints?

No... I would expect the hashes to be the same for the same input, otherwise what use is it? Are you using the same IV on both sides?

[Wikipedia](https://en.wikipedia.org/wiki/BLAKE_(hash_function)) has some example hashes for empty string data, do you get the same hash for empty an string?

squonk11:
In Wikipedia I do not find a blake2b hash (with digest size of 32 bytes) for an empty string. 
But if I calculate a blake2b hash (with digest size of 32 bytes) in Python and with libsodium I get the same results. But I don't get the same hashes for the transmitted files. So probably there is some other issue with my code. I need to check in more detail during the next days.

squonk11:
After modifying my Python code a bit and using an empty string as `Key`, the hashes on both sides seem to be the same. Unfortunately it is not quite clear to my why the previous version did not work ...
Currently I am comparing the hashes on both sides manually. The next step is to send them as JSON object back to the Python script. Do you have a hint how I can send a JSON object as reply? Do I have to send it as `responses::StringResponse`?

squonk11:
Now I have the first version of the test ready.
I used the StringResponse to reply with the JSON string. In Python this string gets converted to a Python dict. The hashes of the files are then compared to the hashes returned from the HTTP server and a success or error message is printed.
I tested several times with the 1000 icon files and no error ocured. The only message I see is again that the watchdog on one CPU is triggered due to not being serviced. But the transmission still completes.
Shall I upload this as first version for discussion into this PR since it is still open?

PerMalmberg:
Yes please do that. Did you try with larger files too?

squonk11:
yes, I tried also with lager files; e.g. three files: 8MB, 10MB and 1kB
I will provide the update soon.

PerMalmberg:
I tried to run this in as a Linux version, but that unfortunately results in a crash.

`[1]    31201 abort (core dumped)  ./http_files_upload_test.elf`

I'll have to debug this and see what goes wrong.

squonk11:
Dies it crash during or after file(s) Transmission? 

Am 12. Juni 2020 22:00:49 MESZ schrieb Per Malmberg <notifications@github.com>:
>I tried to run this in as a Linux version, but that unfortunately
>results in a crash.
>
>`[1]    31201 abort (core dumped)  ./http_files_upload_test.elf`
>
>I'll have to debug this and see what goes wrong.
>
>-- 
>You are receiving this because you were mentioned.
>Reply to this email directly or view it on GitHub:
>https://github.com/PerMalmberg/Smooth/pull/132#issuecomment-643459499

-- 
Diese Nachricht wurde von meinem Android-Gerät mit K-9 Mail gesendet.

PerMalmberg:
As soon as transmission starts. Fist file.

squonk11:
I checked the sources again. Maybe there could be two weak points:

1. in UploadResponder::start_of_request() I am using `sodium_malloc()`. Maybe a stub is needed for this?
2. In MIMEParser::parse() I define status as static. Maybe it must be a private member variable of MIMEParser?

Are you using the secure or the insecure server for the test?

PerMalmberg:
I used the insecure server. It's most likely related to the sodium stuff, not sure why though. Adding gdb to the container right now to debug it.

A static local in parse() is no good since it then is is shared between instances.

PerMalmberg:
First crash solved - sodium_init() must be called prior to using the library.

Another problem that then immediately showed it self - in `end_of_request()`, the call to `free(pState)` must be changed to `sodium_free()`.

This clearly shows why it is beneficial to perform most of the work on a Linux system - the debugging facilities are so much better.

I'll be testing this a bit and will push the changes to you branch (if I can).

PerMalmberg:
I'm done testing now, just waiting for the CI scripts to run locally. I see that you changed it to a local variable. I'll merge that into my local copy and push my changes to this PR.

squonk11:
o.k. thanks. Then I will check again.
Sure, testing under a "real" OS is much easier because you have all tools available. Usually I am working on µCs where this is not possible - even the logging to a terminal (like here in esp-idf) is not possible. Then debugging becomes very ugly...

PerMalmberg:
Just pushed my changes, have a look.

Yes, uCs are a pain to debug, which is why I made Smooth so that it would be possible to use all the cool toys available :)

squonk11:
yes, great idea!
Do you know if I just have to do `git pull` in order to receive your changes?

PerMalmberg:
Yes a pull should be all you need.

squonk11:
I tested it again and it seems to work - after adjusting the url in upload.py. 
Interesting to see your modifications in upload.py - for me it was more or less the first Python script I made.
Are there any other changes needed from my side? Maybe adding `PR132` to `CONTRIBUTORS.md`?

PerMalmberg:
Yes, please add it to the contributions file, then I'll accept this PR.

PerMalmberg:
Merged! Thank you for your contribution @squonk11, great work.

squonk11:
Thank you also to you - without your support things might have been much more difficult. 
BTW: the last change to CONTRIBUTORS.md got lost somehow...


Am 13. Juni 2020 19:57:25 MESZ schrieb Per Malmberg <notifications@github.com>:
>Merged! Thank you for your contribution @squonk11, great work.
>
>-- 
>You are receiving this because you were mentioned.
>Reply to this email directly or view it on GitHub:
>https://github.com/PerMalmberg/Smooth/pull/132#issuecomment-643656797

-- 
Diese Nachricht wurde von meinem Android-Gerät mit K-9 Mail gesendet.

PerMalmberg:
Huh, how odd. I'll fix that.

PerMalmberg:
@squonk11 Dunno if the commit was advertised to you, but it was just a typo. It's fixed in master now.

PR #133, opened 2020-08-08
Added missing <cstdint> include for uint8_t

COM8:
This PR adds the missing include for `uint8_t`.
Without it derived classes do not compile.
```log
error: 'uint8_t' does not name a type
```


PerMalmberg:
Cool, thank you.

PR #136, opened 2020-10-17
Add support for wired ethernet

elcojacobs:
First of all, your framework looks well designed so I think this could be my first of many PR's.

I added support for wired Ethernet, which required a refactor of the WiFi class too.
Please let me know your thoughts on the changes.

I tested this on custom hardware with the internal MAC and a LAN8742 phy.
The LAN8742 is very similar to the 8720, but not included yet in esp-idf. Will later send a PR for that to esp-idf.

**Changes**
- Added a base class for a network interface. Moved managing hostname, interface, and addresses to this base class. The network class counts the number of interfaces because the network stack should only be initialized once per application.
- Changed how host_name is stored. esp_idf only copies a pointer, so string.c_str() can become invalid if the string is reallocated. host_name should also be set before the interface is started. I chose to store the hostname in each interface, so they can have a unique hostname. An alternative is to make the hostname static and use the same for both wifi and ethernet. Not sure what is desired here.
- Added Ethernet class. Code is based on the ethernet basic example from esp-idf.
- Added ethernet settings to kconfig
- Added copy of socket_server_test that also uses ethernet.

Result of test app and idf.py monitor below.
You can see that the ESP gets 2 ip addresses. Connections are accepted on both IPs, tested with:
```
echo ` date` | nc 192.168.1.115 8080 -w1
echo ` date` | nc 192.168.1.116 8080 -w1
```

<details>
<pre>
entry 0x40080688
I (29) boot: ESP-IDF v4.2-beta1-87-gb83aa9fda 2nd stage bootloader
I (29) boot: compile time 18:26:28
I (29) boot: chip revision: 1
I (33) boot_comm: chip revision: 1, min. bootloader chip revision: 0
I (40) boot.esp32: SPI Speed      : 40MHz
I (45) boot.esp32: SPI Mode       : DIO
I (50) boot.esp32: SPI Flash Size : 4MB
I (54) boot: Enabling RNG early entropy source...
I (60) boot: Partition Table:
I (63) boot: ## Label            Usage          Type ST Offset   Length
I (70) boot:  0 nvs              WiFi data        01 02 00009000 00006000
I (78) boot:  1 phy_init         RF data          01 01 0000f000 00001000
I (85) boot:  2 factory          factory app      00 00 00010000 00300000
I (93) boot:  3 app_storage      Unknown data     01 81 00310000 00084000
I (100) boot: End of partition table
I (105) boot_comm: chip revision: 1, min. application chip revision: 0
I (112) esp_image: segment 0: paddr=0x00010020 vaddr=0x3f400020 size=0x3b2ec (242412) map
I (213) esp_image: segment 1: paddr=0x0004b314 vaddr=0x3ffb0000 size=0x03dbc ( 15804) load
I (220) esp_image: segment 2: paddr=0x0004f0d8 vaddr=0x40080000 size=0x00404 (  1028) load
0x40080000: _WindowOverflow4 at /home/elco/repos/esp-idf/components/freertos/xtensa/xtensa_vectors.S:1730

I (221) esp_image: segment 3: paddr=0x0004f4e4 vaddr=0x40080404 size=0x00b34 (  2868) load
I (230) esp_image: segment 4: paddr=0x00050020 vaddr=0x400d0020 size=0xc4214 (803348) map
0x400d0020: _stext at ??:?

I (543) esp_image: segment 5: paddr=0x0011423c vaddr=0x40080f38 size=0x154fc ( 87292) load
I (594) boot: Loaded app from partition at offset 0x10000
I (594) boot: Disabling RNG early entropy source...
I (594) cpu_start: Pro cpu up.
I (598) cpu_start: Application information:
I (603) cpu_start: Project name:     server_socket_ethernet_test
I (609) cpu_start: App version:      9de844f-dirty
I (615) cpu_start: Compile time:     Oct 17 2020 18:38:57
I (621) cpu_start: ELF file SHA256:  d997a84045d58fce...
I (627) cpu_start: ESP-IDF:          v4.2-beta1-87-gb83aa9fda
I (633) cpu_start: Starting app cpu, entry point is 0x400818a0
0x400818a0: call_start_cpu1 at /home/elco/repos/esp-idf/components/esp32/cpu_start.c:282

I (0) cpu_start: App cpu up.
I (644) heap_init: Initializing. RAM available for dynamic allocation:
I (651) heap_init: At 3FFAE6E0 len 00001920 (6 KiB): DRAM
I (657) heap_init: At 3FFB9D90 len 00026270 (152 KiB): DRAM
I (663) heap_init: At 3FFE0440 len 00003AE0 (14 KiB): D/IRAM
I (669) heap_init: At 3FFE4350 len 0001BCB0 (111 KiB): D/IRAM
I (676) heap_init: At 40096434 len 00009BCC (38 KiB): IRAM
I (682) cpu_start: Pro cpu start user code
I (729) spi_flash: detected chip: gd
I (730) spi_flash: flash io: dio
W (730) spi_flash: Detected size(16384k) larger than the size in the binary image header(4096k). Using the size in the binary image header.
I (740) cpu_start: Starting scheduler on PRO CPU.
I (0) cpu_start: Starting scheduler on APP CPU.
I (812) FSLock: Max number of open files set to: 5
I (825) system_api: Base MAC address is not set
I (825) system_api: read default base MAC address from EFUSE
I (838) App::Init: Starting Ethernet...
I (839) esp_eth.netif.glue: 24:62:ab:c3:8b:bf
I (839) esp_eth.netif.glue: ethernet attached to netif
I (853) App::Init: Starting wifi...
I (857) wifi:wifi driver task: 3ffcf9f0, prio:23, stack:3584, core=0
I (873) wifi:wifi firmware version: d3be909
I (873) wifi:wifi certification version: v7.0
I (873) wifi:config NVS flash: enabled
I (873) wifi:config nano formating: disabled
I (878) wifi:Init data frame dynamic rx buffer num: 32
I (882) wifi:Init management frame dynamic rx buffer num: 32
I (888) wifi:Init management short buffer num: 32
I (892) wifi:Init static tx buffer num: 16
I (896) wifi:Init static rx buffer size: 1600
I (900) wifi:Init static rx buffer num: 10
I (904) wifi:Init dynamic rx buffer num: 32
I (909) wifi_init: rx ba win: 16
I (911) wifi_init: tcpip mbox: 32
I (915) wifi_init: udp mbox: 6
I (919) wifi_init: tcp mbox: 6
I (923) wifi_init: tcp tx win: 5744
I (927) wifi_init: tcp rx win: 5744
I (931) wifi_init: tcp mss: 1440
I (935) wifi_init: WiFi IRAM OP enabled
I (940) wifi_init: WiFi RX IRAM OP enabled
I (1027) phy: phy_version: 4390, 6b3c1f2, Sep 10 2020, 15:09:07, 0, 0
I (1029) wifi:mode : sta (24:62:ab:c3:8b:bc)
I (1037) SocketDispatcher: Active sockets: 0
I (1637) wifi:new:<5,0>, old:<1,0>, ap:<255,255>, sta:<5,0>, prof:1
I (2173) wifi:state: init -> auth (b0)
I (2178) wifi:state: auth -> assoc (0)
I (2184) wifi:state: assoc -> run (10)
I (2211) wifi:connected with BrouwerijDeSchutter, aid = 3, channel 5, BW20, bssid = 7c:ff:4d:16:65:fa
I (2212) wifi:security: WPA2-PSK, phy: bg, rssi: -76
I (2215) wifi:pm start, type: 1

I (2352) wifi:AP's beacon interval = 102400 us, DTIM period = 1
I (3316) SocketDispatcher: Network up, sockets will be restarted.
I (3317) esp_netif_handlers: sta ip: 192.168.1.116, mask: 255.255.255.0, gw: 192.168.1.1
I (3318) SocketDispatcher: Network up, sockets will be restarted.
I (3330) Socket: [0.0.0.0, 8080, 54, 0x3ffdb9c0]: Created server socket
I (3815) esp_netif_handlers: eth ip: 192.168.1.115, mask: 255.255.255.0, gw: 192.168.1.1
I (3822) SocketDispatcher: Network up, sockets will be restarted.
I (3823) SocketDispatcher: Network up, sockets will be restarted.
I (3828) Socket: [0.0.0.0, 8080, 54, 0x3ffdb9c0]: Server stopping
E (3835) SocketDispatcher: Shutdown error: Invalid argument
I (3841) SocketDispatcher: Active sockets: 0
I (3846) SocketDispatcher: Active sockets: 0
I (3851) Socket: [0.0.0.0, 8080, 54, 0x3ffdb9c0]: Created server socket
I (13282) ServerSocket: Connection accepted
I (13284) SocketDispatcher: Active sockets: 2
I (14283) Socket: [, 0, 55, 0x3ffdd744]: Underlying socket closed (recv returned 0)
I (14283) Socket: [, 0, 55, 0x3ffdd744]: Socket stopping
I (14289) Socket: [, 0, -1, 0x3ffdd744]: Disconnected
I (14293) SocketDispatcher: Active sockets: 1
I (18954) ServerSocket: Connection accepted
I (18956) SocketDispatcher: Active sockets: 2
I (19971) Socket: [, 0, 55, 0x3ffdd744]: Underlying socket closed (recv returned 0)
I (19971) Socket: [, 0, 55, 0x3ffdd744]: Socket stopping
I (19977) Socket: [, 0, -1, 0x3ffdd744]: Disconnected
I (19980) SocketDispatcher: Active sockets: 1
</pre>
</details>

**Formatting**
For my own projects, I use clang-format. Integrates very nicely with vscode to format on save.
I ran uncrustify for this PR, because my clang-format rules are different.

I use these settings and the `xaver.clang-format` vscode plugin:
https://github.com/BrewBlox/brewblox-firmware/blob/develop/.clang-format

Have you tried/considered clang-format instead of uncrustify?

### DIFF
--- a/lib/smooth/core/network/Ethernet.cpp
+++ b/lib/smooth/core/network/Ethernet.cpp
@@ -40,48 +40,13 @@ using namespace smooth::core;
 
 namespace smooth::core::network {
 
-Ethernet::Ethernet(std::string name)
+Ethernet::Ethernet(std::string&& name)
     : NetworkInterface(std::move(name))
 {
-}
-
-Ethernet::~Ethernet()
-{
-    deinit();
-}
-
-bool
-Ethernet::init()
-{
-    esp_err_t err = do_init();
-    if (err != ESP_OK) {
-        Log::error(interface_name, "init", esp_err_to_name(err));
-        return false;
-    }
-    return true;
-}
-
-void
-Ethernet::deinit()
-{
-    esp_event_handler_instance_unregister(IP_EVENT, ESP_EVENT_ANY_ID, instance_ip_event);
-    esp_event_handler_instance_unregister(ETH_EVENT, ESP_EVENT_ANY_ID, instance_eth_event);
-    if (eth_handle) {
-        esp_eth_stop(eth_handle);
-    }
-}
-
-esp_err_t
-Ethernet::do_init()
-{
-    esp_err_t err = ESP_OK;
-
     esp_netif_config_t cfg = ESP_NETIF_DEFAULT_ETH();
     interface = esp_netif_new(&cfg);
-    err = esp_eth_set_default_handlers(interface);
-    if (err != ESP_OK) {
-        return err;
-    }
+    esp_eth_set_default_handlers(interface);
+
     esp_event_handler_instance_register(ETH_EVENT,
                                         ESP_EVENT_ANY_ID,
                                         &Ethernet::eth_event_callback,
@@ -96,37 +61,38 @@ Ethernet::do_init()
 
     eth_mac_config_t mac_config = ETH_MAC_DEFAULT_CONFIG();
     eth_phy_config_t phy_config = ETH_PHY_DEFAULT_CONFIG();
-
-    phy_config.phy_addr = CONFIG_EXAMPLE_ETH_PHY_ADDR;
-    phy_config.reset_gpio_num = CONFIG_EXAMPLE_ETH_PHY_RST_GPIO;
-#if CONFIG_EXAMPLE_USE_INTERNAL_ETHERNET
-    mac_config.smi_mdc_gpio_num = CONFIG_EXAMPLE_ETH_MDC_GPIO;
-    mac_config.smi_mdio_gpio_num = CONFIG_EXAMPLE_ETH_MDIO_GPIO;
+    phy_config.phy_addr = CONFIG_SMOOTH_ETH_PHY_ADDR;
+    phy_config.reset_gpio_num = CONFIG_SMOOTH_ETH_PHY_RST_GPIO;
+#if CONFIG_SMOOTH_USE_INTERNAL_ETHERNET
+    mac_config.smi_mdc_gpio_num = CONFIG_SMOOTH_ETH_MDC_GPIO;
+    mac_config.smi_mdio_gpio_num = CONFIG_SMOOTH_ETH_MDIO_GPIO;
     mac = esp_eth_mac_new_esp32(&mac_config);
-#if CONFIG_EXAMPLE_ETH_PHY_IP101
+#if CONFIG_SMOOTH_ETH_PHY_IP101
     phy = esp_eth_phy_new_ip101(&phy_config);
-#elif CONFIG_EXAMPLE_ETH_PHY_RTL8201
+#elif CONFIG_SMOOTH_ETH_PHY_RTL8201
     phy = esp_eth_phy_new_rtl8201(&phy_config);
-#elif CONFIG_EXAMPLE_ETH_PHY_LAN8720
+#elif CONFIG_SMOOTH_ETH_PHY_LAN8720
     phy = esp_eth_phy_new_lan8720(&phy_config);
-#elif CONFIG_EXAMPLE_ETH_PHY_LAN8742
+#elif CONFIG_SMOOTH_ETH_PHY_LAN8742
     phy = esp_eth_phy_new_lan8742(&phy_config);
-#elif CONFIG_EXAMPLE_ETH_PHY_DP83848
+#elif CONFIG_SMOOTH_ETH_PHY_DP83848
     phy = esp_eth_phy_new_dp83848(&phy_config);
-#elif CONFIG_EXAMPLE_ETH_PHY_KSZ8041
+#elif CONFIG_SMOOTH_ETH_PHY_KSZ8041
     phy = esp_eth_phy_new_ksz8041(&phy_config);
+#else
+#error CONFIG_SMOOTH_ETH_PHY_type not set
 #endif
-#elif CONFIG_EXAMPLE_USE_DM9051
+#elif CONFIG_SMOOTH_USE_DM9051
     gpio_install_isr_service(0);
     spi_device_handle_t spi_handle = NULL;
     spi_bus_config_t buscfg = {
-        .miso_io_num = CONFIG_EXAMPLE_DM9051_MISO_GPIO,
-        .mosi_io_num = CONFIG_EXAMPLE_DM9051_MOSI_GPIO,
-        .sclk_io_num = CONFIG_EXAMPLE_DM9051_SCLK_GPIO,
+        .miso_io_num = CONFIG_SMOOTH_DM9051_MISO_GPIO,
+        .mosi_io_num = CONFIG_SMOOTH_DM9051_MOSI_GPIO,
+        .sclk_io_num = CONFIG_SMOOTH_DM9051_SCLK_GPIO,
         .quadwp_io_num = -1,
         .quadhd_io_num = -1,
     };
-    err = spi_bus_initialize(CONFIG_EXAMPLE_DM9051_SPI_HOST, &buscfg, 1);
+    spi_bus_initialize(CONFIG_SMOOTH_DM9051_SPI_HOST, &buscfg, 1);
     if (err != ESP_OK) {
         return err;
     }
@@ -134,36 +100,47 @@ Ethernet::do_init()
         .command_bits = 1,
         .address_bits = 7,
         .mode = 0,
-        .clock_speed_hz = CONFIG_EXAMPLE_DM9051_SPI_CLOCK_MHZ * 1000 * 1000,
-        .spics_io_num = CONFIG_EXAMPLE_DM9051_CS_GPIO,
+        .clock_speed_hz = CONFIG_SMOOTH_DM9051_SPI_CLOCK_MHZ * 1000 * 1000,
+        .spics_io_num = CONFIG_SMOOTH_DM9051_CS_GPIO,
         .queue_size = 20};
-    err = spi_bus_add_device(CONFIG_EXAMPLE_DM9051_SPI_HOST, &devcfg, &spi_handle);
+    spi_bus_add_device(CONFIG_SMOOTH_DM9051_SPI_HOST, &devcfg, &spi_handle);
     if (err != ESP_OK) {
         return err;
     }
     /* dm9051 ethernet driver is based on spi driver */
     eth_dm9051_config_t dm9051_config = ETH_DM9051_DEFAULT_CONFIG(spi_handle);
-    dm9051_config.int_gpio_num = CONFIG_EXAMPLE_DM9051_INT_GPIO;
+    dm9051_config.int_gpio_num = CONFIG_SMOOTH_DM9051_INT_GPIO;
     mac = esp_eth_mac_new_dm9051(&dm9051_config, &mac_config);
     phy = esp_eth_phy_new_dm9051(&phy_config);
+#else
+#error CONFIG_SMOOTH_USE_INTERNAL_ETHERNET or CONFIG_SMOOTH_USE_DM9051 should be set
 #endif
     esp_eth_config_t config = ETH_DEFAULT_CONFIG(mac, phy);
-    err = esp_eth_driver_install(&config, &eth_handle);
-    if (err != ESP_OK) {
-        return err;
+    esp_eth_driver_install(&config, &eth_handle);
+}
+
+Ethernet::~Ethernet()
+{
+    esp_event_handler_instance_unregister(IP_EVENT, ESP_EVENT_ANY_ID, instance_ip_event);
+    esp_event_handler_instance_unregister(ETH_EVENT, ESP_EVENT_ANY_ID, instance_eth_event);
+    if (eth_handle) {
+        esp_eth_stop(eth_handle);
     }
-    /* attach Ethernet driver to TCP/IP stack */
-    err = esp_netif_attach(interface, esp_eth_new_netif_glue(eth_handle));
-    if (err != ESP_OK) {
-        return err;
-    }
-    /* start Ethernet driver state machine */
-    err = esp_eth_start(eth_handle);
-    if (err != ESP_OK) {
-        return err;
+}
+
+void
+Ethernet::start()
+{
+    apply_host_name();
+    auto err = esp_netif_attach(interface, esp_eth_new_netif_glue(eth_handle));
+    if (err == ESP_OK) {
+        /* start Ethernet driver state machine */
+        err = esp_eth_start(eth_handle);
     }
 
-    return err;
+    if (err != ESP_OK) {
+        Log::error(interface_name, "start", esp_err_to_name(err));
+    }
 }
 
 bool
--- a/lib/smooth/core/network/Ethernet.cpp
+++ b/lib/smooth/core/network/Ethernet.cpp
@@ -1,0 +1,226 @@
+/*
+Smooth - A C++ framework for embedded programming on top of Espressif's ESP-IDF
+Copyright 2019 Per Malmberg (https://gitbub.com/PerMalmberg)
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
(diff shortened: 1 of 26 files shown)


elcojacobs:
Test build is failing because no stubs for esp_eth exist in mock-idf.
I can add some empty functions, but not sure if that's the right approach.

The mocks here are mostly just empty stubs. On my current platform, we can run our entire app as a cross compiled executable, including network, GPIO mocking, etc. Would be nice to have this on ESP32 too.

I found this repo, but haven't tried it yet.
https://github.com/mireq/esp32-simulator

PerMalmberg:
Smooth can be cross compiled and network was functional when running on Linux prior to this PR so that needs to work. IO is currently just mocked.

I'll try to find some time tomorrow or at least the next few days to look at this PR.

I have considered clang format, but found uncrustify worked better, at least it did a few years ago.

elcojacobs:
I added `eth` stubs in mock-idf.

Compile now fails on acces_point test, because it still expects to have some static functions on the Wifi class, that I have moved to the NetworkInterface class.

It does make sense to have only one Wifi instance.
In theory, you can have multiple Ethernet instances on the SPI bus. But we can choose not to implement this use case.

An option is a global wifi getter and construct it on first use. It can then be removed from App members.
`
 Wifi& get_wifi(){
     static Wifi* wifi = WiFi();
     return *wifi;
 }
`

elcojacobs:
Global singleton added for Wifi, tests passing :smile:

PerMalmberg:
Had a quick look, overall it looks good. I have a few things I want to think about/look closer at in an actual code-editor.

It's a bit outside the coding-part, but I feel like a small documentation section about the h/w side could be of use for the wired ethernet. Perhaps some links to resources that gets people started?

elcojacobs:
I would just refer to the esp-idf docs, because the options I added to kconfig are copy-paste from their examples.
The only change I made is to set the default to phy address 0. That gotcha took me a while to figure out before I got it to work.
I am also using the framework functions to initialize everything, nothing custom about that. Just some C++ around the api calls.

For reference, I am using the internal MAC with a LAN8742 phy. It is cheaper and easier to get than the 8720.
I let the ESP generate the 50Mhz clock for the phy on pin gpio0, and use the default pins.
I use the WROOM-32D. If I remember correctly, the variants with external PSRAM don't have ethernet.
Ethernet needs quite a lot of dedicated pins.

```
CONFIG_SMOOTH_USE_INTERNAL_ETHERNET=y
# CONFIG_SMOOTH_USE_DM9051 is not set
# CONFIG_SMOOTH_ETH_PHY_IP101 is not set
# CONFIG_SMOOTH_ETH_PHY_RTL8201 is not set
# CONFIG_SMOOTH_ETH_PHY_LAN8720 is not set
CONFIG_SMOOTH_ETH_PHY_LAN8742=y
# CONFIG_SMOOTH_ETH_PHY_DP83848 is not set
CONFIG_SMOOTH_ETH_MDC_GPIO=23
CONFIG_SMOOTH_ETH_MDIO_GPIO=18
CONFIG_SMOOTH_ETH_PHY_RST_GPIO=5
CONFIG_SMOOTH_ETH_PHY_ADDR=0
```

No need in duplicating the docs here:
https://docs.espressif.com/projects/esp-idf/en/latest/esp32/hw-reference/esp32/get-started-ethernet-kit.html#gpio-allocation
https://docs.espressif.com/projects/esp-idf/en/latest/esp32/api-reference/network/esp_eth.html

I would add those links and leave it at that.

lib/smooth/include/smooth/core/network/Wifi.h
```diff
@@ -94,15 +81,14 @@ namespace smooth::core::network
...
             std::string password{};
-            static struct esp_ip4_addr ip;
 
             esp_netif_t* interface{ nullptr };
             esp_event_handler_instance_t instance_wifi_event{};
             esp_event_handler_instance_t instance_ip_event{};
     };
+
+    Wifi&
+    get_wifi();
```
PerMalmberg, line 93:
```suggestion
    Wifi& get_wifi();
```

test/server_socket_ethernet_test/StreamPacket.h
```diff
@@ -0,0 +1,54 @@
...
+        }
+
+        std::array<uint8_t, 1>& data()
+        {
+            return buff;
+        }
+
+    private:
+        std::array<uint8_t, 1> buff{};
+};
```
PerMalmberg, line 54: The ethernet test looks like a copy of the Wifi test. Can we do something better than to duplicate the entire code base for the test so that future improvements don't have to be made for both?

test/server_socket_ethernet_test/server_socket_ethernet_test.cpp
```diff
@@ -0,0 +1,67 @@
...
+        network::Wifi& wifi = get_wifi();
+        wifi.set_host_name("smooth_wifi");
+        wifi.set_auto_connect(true);
+        wifi.set_ap_credentials(WIFI_SSID, WIFI_PASSWORD);
+        wifi.connect_to_ap();
+
+        // The server creates StreamingClients which are self-sufficient and never seen by the main
+        // application (unless the implementor adds such bindings).
+        server = ServerSocket<StreamingClient, StreamingProtocol, void>::create(*this, 5, 5);
+        server->start(std::make_shared<IPv4>("0.0.0.0", 8080));
```
PerMalmberg, line 62: Nos that there are two interfaces, which interface will the sockets bind to? Both?

lib/smooth/include/smooth/core/network/NetworkInterface.h
```diff
@@ -0,0 +1,66 @@
...
+            bool get_local_mac_address(std::array<uint8_t, 6>& m) const;
+
+        protected:
+            void apply_host_name();
+
+            esp_ip4_addr ip = { 0 };
+            bool connected = false;
+            esp_netif_t* interface{ nullptr };
+            static uint8_t interface_count;
+            std::string interface_name;
```
PerMalmberg, line 60:
```suggestion
            esp_ip4_addr ip = {};
            bool connected = false;
            esp_netif_t* interface{};
            static uint8_t interface_count[};
            std::string interface_name[};
```
Lets use C++11's zero-initialization for consistency, I've tried to do that elsewhere.

lib/smooth/core/network/Wifi.cpp
```diff
@@ -119,7 +109,7 @@ namespace smooth::core::network
 
     bool Wifi::is_connected_to_ap() const
     {
-        return connected_to_ap;
+        return connected;
```
PerMalmberg, line 112: Now that Wifi inherits from `NetworkInterface` this method is obsolete and replaced by is_connected() so lets remove it.

PerMalmberg (changes requested):
Left several comments throughout the code.

I take it you haven't tried all the different network interfaces considering I'm 99.9% sure that CONFIG_SMOOTH_USE_DM9051  won't compile. If this is the case please remove the parts that you haven't actually run. There's a lot going on in the Ethernet constructor, perhaps move it into a reentrant-safe setup method called from start()?

Although it is nice to have all the options available, if the code isn't tested I don't want to merge it.

> I would add those links and leave it at that.

Fair enough.

lib/smooth/core/network/NetworkInterface.cpp
```diff
@@ -0,0 +1,97 @@
+#include "smooth/core/network/NetworkInterface.h"
+#include "smooth/core/logging/log.h"
+#include <esp_err.h>
+#include <esp_netif.h>
+#include <sstream>
+
+using namespace smooth::core::logging;
+
+namespace smooth::core::network {
+    uint8_t NetworkInterface::interface_count = 0;
```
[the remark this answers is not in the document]
elcojacobs: It should be static in the NetworkInterface class. 
Its sole purpose is to know if netif is initialized. This is needed for both WiFi and Ethernet, which is why it is in the shared class.

The problem was just that it was in Wifi init, under the assumption that it was the only network interface.

lib/smooth/core/network/Ethernet.cpp
```diff
@@ -0,0 +1,223 @@
...
+            .sclk_io_num = CONFIG_SMOOTH_DM9051_SCLK_GPIO,
+            .quadwp_io_num = -1,
+            .quadhd_io_num = -1,
+        };
+        spi_bus_initialize(CONFIG_SMOOTH_DM9051_SPI_HOST, &buscfg, 1);
+
+        if (err != ESP_OK)
+        {
+            return err;
+        }
```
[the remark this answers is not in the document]
elcojacobs: No, I don't have the hardware. 
You are right. It is copy-pasted from the example code which was not in a constructor.

lib/smooth/core/network/Ethernet.cpp
```diff
@@ -0,0 +1,223 @@
...
+                eth->ip.addr = 0;
+                eth->connected = false;
+                publish_status(eth->connected, true);
+            }
+            break;
+            }
+        }
+        else if (event_base == IP_EVENT)
+        {
+            if (event_id == IP_EVENT_STA_GOT_IP
```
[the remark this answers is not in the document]
elcojacobs: good catch. I'll check. This code is for both wifi and ethernet, but I don't see IP events defined for ethernet.

https://github.com/espressif/esp-idf/blob/fb3d648e07033b952578f187b4fcbf7fd8aced3b/examples/common_components/protocol_examples_common/connect.c

lib/smooth/include/smooth/core/network/Wifi.h
elcojacobs: yes, changed my clang-format to not do this.

test/server_socket_ethernet_test/StreamPacket.h
elcojacobs: I copy-pasted to not affect your test.
Ethernet could be included based on #ifdefs by kconfig in the existing test too.

test/server_socket_ethernet_test/server_socket_ethernet_test.cpp
elcojacobs: yes, it all goes through global netif.

lib/smooth/include/smooth/core/network/NetworkInterface.h
elcojacobs: I assume `[}` is a typo.

I think nullptr and {0} for the array are more explicit.
I actually didn't know that for types without constructor, {} will default initialize to zero.

interface_name is set in the constructor, and strings are default initialized to empty, so I am not sure I agree with you for the string.

elcojacobs:
Thanks for the feedback.

After playing with smooth for a bit, I actually think it might be a bit too opinionated for my goals.
It uses a lot of events by default and is quite verbose, with ease of use over performance.
Many of the logs generated by smooth are duplications of the logs already emitted by esp-idf.

I have been playing with asio, and I think it will be a better fit for my application.
I prefer to stay a bit closer to esp-idf where possible, and use the framework provided implementations for mdns for example.

I'll still make the changes to this PR and will look at smooth for inspiration here and there, but I think it will not combine well with asio. asio is not even available when smooth is included, I am not sure why.

peterus:
@elcojacobs thanks for the implementation of the ethernet part! It will help on my current project a lot!
please finish your current PR if possible, if you need/want help: just give me a ping ;)

elcojacobs:
I don't have time to work on this in the coming days. Feel free to pull my branch and work on it or even finish it.

PerMalmberg:
> After playing with smooth for a bit, I actually think it might be a bit too opinionated for my goals.
> It uses a lot of events by default and is quite verbose, with ease of use over performance.

That's is a correct observation; it's was one of my my goal with it.

> Many of the logs generated by smooth are duplications of the logs already emitted by esp-idf.

Oh, anything in particular you have in mind here?

> I have been playing with asio, and I think it will be a better fit for my application.
I prefer to stay a bit closer to esp-idf where possible, and use the framework provided implementations for mdns for example.

Then you should ofc use it! The right tool for the job etc. :)

> I'll still make the changes to this PR and will look at smooth for inspiration here and there, but I think it will not combine well with asio. asio is not even available when smooth is included, I am not sure why.

Apart from a possible conflict with the network layer I don't see a problem using asio together with smooth. I've not disabled asio in Smooth afaik.

PR #140, opened 2021-01-09
added some more mime types to 'get_content_type()' in http_utils.cpp

squonk11:
I added some more mime types to 'get_content_type()' in http_utils.cpp. 
Additionally I added `#pragma GCC diagnostic ignored "-Wold-style-cast"` to some header files because the actual master of esp-idf has some issues with old-style-cast.

### DIFF
--- a/lib/smooth/application/network/http/http_utils.cpp
+++ b/lib/smooth/application/network/http/http_utils.cpp
@@ -97,6 +97,34 @@ namespace smooth::application::network::http::utils
         {
             return "text/html";
         }
+        else if (ext == ".css")
+        {
+            return "text/css";
+        }
+        else if (ext == ".js")
+        {
+            return "text/javascript";
+        }
+        else if (ext == ".svg")
+        {
+            return "image/svg+xml";
+        }
+        else if (ext == ".gif")
+        {
+            return "image/gif";
+        }
+        else if (ext == ".json")
+        {
+            return "application/json";
+        }
+        else if (ext == ".mp3")
+        {
+            return "audio/mpeg";
+        }
+        else if (ext == ".ttf")
+        {
+            return "font/ttf";
+        }
 
         return "application/octet-stream";
     }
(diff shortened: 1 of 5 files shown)


PerMalmberg:
Does the filesystem support case? I.e. is `image.svg` the same as `image.SVG` ? If yes, `ext` should be transformed to lower-case before the comparisons. If not, I'm happy to merge this.

squonk11:
Shouldn't it be vice versa: if the filesystem does **not** support case `ext` must be transformed to lower-case? Because: if the filesystem does not distinguish case, `.SVG` **and** `.svg` must always match the comparison.

PerMalmberg:
Well, depends on if files always are stored in lower case or not? I don't
remember what the case is for the fs used in IDF.

PerMalmberg:
> master of esp-idf has some issues with old-style-cast.

Maybe it is time to rebuild the docker image used in CI, it is a bit dated at this time.

squonk11:
I tested the case sensitivity of the file system and it seems that it is **not** case-sensitive. Additionally I found this [discussion](https://esp32.com/viewtopic.php?t=5413) where people also say that it is **case-insensitive.** That seems to be logical as FAT is a filesystem coming from DOS/Windows where files are always case-insensitive. 
But nevertheless if a user uses `href="css/style.CSS"` he'd probably expect that this will match the file `style.css`. In order to achieve this `ext` would need to be converted to lowercase. Unfortunately this has some drawbacks:

- there is no toupper() method in std::string. This can be circumvented using something like `for_each(ext.begin(), ext.end(), [](char& in){ in = ::tolower(in); });`
- ext currently is defined as `const auto&`. So, I would need to make a local copy before comparing

Shall I do this?

Concerning updating the docker image: unfortunately I am no familiar with Docker. So, it would be quite cumbersome for me to do it...

PerMalmberg:
Please do. Smooth already has a to_lower_copy: https://github.com/PerMalmberg/Smooth/blob/0e277276623a89b7484c8d9205776436ca947502/lib/smooth/include/smooth/core/util/string_util.h#L55

I'll do the Docker update, whenever I get around to it.

PerMalmberg (approved):

PerMalmberg:
Nice 👍

PR #143, opened 2021-03-05
Wifi UPDATES

enelson1001:
Updated to use master and v4.3-beta1 esp-idf
Added three static methods that return strings for ip address, netmask and gateway


enelson1001:
Going to bed - not sure why its failing (the pull_request).  The code compiles and runs on my project.

PerMalmberg:
> Going to bed - not sure why its failing (the pull_request). The code compiles and runs on my project.

Probably because CI runs with an older version of IDF. I'll try to get around to make an update to that.

PerMalmberg:
Didn't have time to finish it before work, but at least I sorted out some issues with the Dockerfile. I'll likely continue with it this afternoon. Was thinking it may be possible to have the image pull the latest master so that all PRs always builds against that.

PerMalmberg:
I've pushed a new docker image built using the master branch of IDF as of today. Sadly it still fails:
https://github.com/PerMalmberg/Smooth/pull/143/checks?check_run_id=2040701351#step:4:191

```
 ../lib/smooth/include/smooth/core/network/Wifi.h:114:20: error: 'esp_netif_ip_info_t' does not name a type; did you mean 'esp_netif_deinit'?
             static esp_netif_ip_info_t ip_info;
                    ^~~~~~~~~~~~~~~~~~~
                    esp_netif_deinit
```

Maybe this file needs to be included? https://github.com/espressif/esp-idf/blob/178b122c145c19e94ac896197a3a4a9d379cd618/components/esp_netif/include/esp_netif_types.h#L104

enelson1001:
Added the include esp_netif_types.h but didn't fix the problem.  I did not need this include "esp_netif_types.h" when I built my project with the forked Smooth library... Strange

squonk11:
in my modified version of Wifi class I am using event information for getting the ip_info structure. In order to do this I used the header `esp_event/private_include/esp_event_internal.h`. But I am not sure if this is correct. because it is named "...event_internal.h". Maybe the header where you found the declaration of the structure is more correct?

PerMalmberg:
Could there be a new setting in sdkconfig that needs to be enabled? CI
builds uses one from 9 months ago.

enelson1001:
In my project sdkconfig I have
CONFIG_ESP_NETIF_TCPIP_ADAPTER_COMPATIBLE_LAYER=y
but in the forked repo it is not set


My sdkconfig
```
#
# ESP NETIF Adapter
#
CONFIG_ESP_NETIF_IP_LOST_TIMER_INTERVAL=120
CONFIG_ESP_NETIF_TCPIP_LWIP=y
# CONFIG_ESP_NETIF_LOOPBACK is not set
CONFIG_ESP_NETIF_TCPIP_ADAPTER_COMPATIBLE_LAYER=y
# end of ESP NETIF Adapter
```
The forked Smooth repo
```
#
# ESP NETIF Adapter
#
CONFIG_ESP_NETIF_IP_LOST_TIMER_INTERVAL=120
CONFIG_ESP_NETIF_TCPIP_LWIP=y
# CONFIG_ESP_NETIF_LOOPBACK is not set
# CONFIG_ESP_NETIF_TCPIP_ADAPTER_COMPATIBLE_LAYER is not set
# end of ESP NETIF Adapter
```

PerMalmberg:
> Maybe the header where you found the declaration of the structure is more correct?

Yeah, don't include files from folders named private.

> CONFIG_ESP_NETIF_TCPIP_ADAPTER_COMPATIBLE_LAYER
According to the docs (https://docs.espressif.com/projects/esp-idf/en/latest/esp32/api-reference/kconfig.html#config-esp-netif-tcpip-adapter-compatible-layer) thats not what we need since the code code fails on the use of netif-function, i.e. we're using the new network interface,

enelson1001:
My project sdkconfig  has a lot of changes compared to sdkconfig in the Smooth fork.  Do you want me to push my sdkconfig?

enelson1001:
1. I tried the sdconfig from my project and did not help.
2. I'm not sure how docker works so question may not be relevant.  Can the folder mock-idf/components be the problem since it only has a folder called tcpip_adapter and does not include a folder called esp_netif?
3. I just doesn't make sense to me why I can compile and run my project with the updated Wifi.cpp and Wifi.h but the ./CI/build_test.sh doesn't work.

enelson1001:
Never mind mock-idf/components has esp_netif.
In mock-idf/components/esp_netif the file esp_netif_types.h is very sparse. Should it include more?

PerMalmberg:
Too late in the night to check now, but I realized now that I didn't check if it was the Linux or esp32 builds that failed. Perhaps it is the mocks that needs some update.

enelson1001:
I will try and update the mock-idf to include what is missing.  One question in Wif.cppp I am using a new method from esp-idf called esp_ip4addr_ntoa() which returns a pointer to char array that hold the address as a string.  So in the mock-idy do I rewrite this method in c++ to perform the same function?

PerMalmberg:
> So in the mock-idy do I rewrite this method in c++ to perform the same function?

Yes, at least enough such that calling it and possibly using the return value, doesn't cause errors or invalid memory access.

enelson1001:
Well I am a lot closer now after major surgery to mock-idf.  This is my last error from running ./CI/build_test.sh 

```
ERROR -> Running host tests ---------------------------------------------------
docker: Error response from daemon: OCI runtime create failed: container_linux.go:370: starting container process caused: exec: "./linux_unit_tests": stat ./linux_unit_tests: no such file or directory: unknown.
```

1. Am I missing a directory if so where do I put it? 
2. Am I missing a script that runs all the test?

PerMalmberg:
> Well I am a lot closer now after major surgery to mock-idf. This is my last error from running ./CI/build_test.sh
> 
> ```
> ERROR -> Running host tests ---------------------------------------------------
> docker: Error response from daemon: OCI runtime create failed: container_linux.go:370: starting container process caused: exec: "./linux_unit_tests": stat ./linux_unit_tests: no such file or directory: unknown.
> ```
> 
>     1. Am I missing a directory if so where do I put it?
> 
>     2. Am I missing a script that runs all the test?

Yes, I can imagine that there were some changes needed to catch up with IDF.

linux_unit_tests is a project which should be built and run in CI, https://github.com/PerMalmberg/Smooth/blob/master/CI/build_test.sh#L34

Are there no errors prior to that line?

enelson1001:
1. There are no prior errors.
2. I am in the Smooth directory and run the following.
```
~/smooth-forked-repo/Smooth$ ./CI/build_test.sh
```
3. What does this line do as the error states there is no such file or directory associated for "./linux_unix_tests"
'''
run_in_docker --workdir "$source_root/build/test/linux_unit_tests" "$docker_image" "./linux_unit_tests"
```
3. I have a Smooth/build/test/linux_unit_tests folder but it is locked and empty.

**Comment by PerMalmberg** [comment]
It tries to run the unit test application it built in a previous step inside the docker container.

**Comment by PerMalmberg** [comment]
I tried running your fork of Smooth, but it seems you haven't pushed the updated idf-mock yet?

**Comment by enelson1001** [comment]
I pushed all my changes to my fork.  Can you try again.

**Comment by enelson1001** [comment]
1. I did not modify .vscode/c_cpp_properties.json for the new compiler.
2. I did not modify sdkconfig but maybe that should be updated?

**Comment by PerMalmberg** [comment]
I've figured you what goes wrong, but not yet why.

For whatever reason the complied binaries are deleted whenever cmake/ninja is run for the next project. This is a new behavior, hence why CI breaks.

**Comment by PerMalmberg** [comment]
Hopefully the change I just made solves it.

**Comment by PerMalmberg** [comment]
Seems it did fix it :)

**Comment by enelson1001** [comment]
Yes, I ran the test also and no errors. Thank YOU

**Comment by enelson1001** [comment]
Next thing to tackle is the SPI and SDCard.  Do you want me to open up another "open issue" to discuss or just continue in this pull request.

**Comment by PerMalmberg** [comment]
> Next thing to tackle is the SPI and SDCard. Do you want me to open up another "open issue" to discuss or just continue in this pull request.

Probably best to keep it separate. Lets get this one cleaned up and merged.

File: lib/smooth/core/network/Wifi.cpp [hunk 1]
```diff
@@ -143,7 +140,12 @@ namespace smooth::core::network
             }
             else if (event_id == WIFI_EVENT_STA_DISCONNECTED)
             {
-                wifi->ip.addr = 0;
+                //wifi->ip.addr = 0;
+                // set ip, netmask and gateway address to 0.0.0.0
```
PerMalmberg, line 144: Lets remove these

lib/smooth/core/network/Wifi.cpp
```diff
@@ -155,12 +157,17 @@ namespace smooth::core::network
             }
             else if (event_id == WIFI_EVENT_AP_START)
             {
-                wifi->ip.addr = 0xC0A80401; // 192.168.4.1
+                //wifi->ip.addr = 0xC0A80401; // 192.168.4.1
```
PerMalmberg, line 160: And this

lib/smooth/core/network/Wifi.cpp
```diff
@@ -155,12 +157,17 @@ namespace smooth::core::network
...
+                wifi->ip_info.ip.addr = 0xC0A80401; // 192.168.4.1
                 publish_status(true, true);
             }
             else if (event_id == WIFI_EVENT_AP_STOP)
             {
-                wifi->ip.addr = 0;
+                //wifi->ip.addr = 0;
```
PerMalmberg, line 166: this too

PerMalmberg (changes requested):

lib/smooth/core/network/Wifi.cpp
enelson1001: Done

lib/smooth/core/network/Wifi.cpp
enelson1001: I believe I have removed all comments in question (and maybe one more)

PerMalmberg (approved):

PR #146, opened 2021-03-19
SPI-Master, SDCard Updates

enelson1001:
Changes to SPI-Master and SDCard using latest esp-idf.
Changes to test programs to run with SPI-Master and SDCard.
Updated mock-idf - also ran uncrustify on mock-idf files from Wifi update(#143), should have done that before sorry.
Tested on M5Stack which has the SDCard and LCD display on same spi-bus.  I do not have hardware to test the MMC card.


lib/smooth/core/filesystem/SDCard.cpp
```diff
@@ -17,56 +17,47 @@ limitations under the License.
...
         }
 
-        FSLock::set_limit(max_file_count);
-
         return initialized;
     }
 
+    // Note: If the SDCard is the only spi-device using spi-host-bus-master then the class calling this
+    // function should also call Master::deinitialize(spi_host_device_t spi_host) to deinitialize the
+    // spi-bus
```
PerMalmberg, line 53: Can we make this automatic using ref a ref-counter?

lib/smooth/core/io/spi/Master.cpp
```diff
@@ -23,63 +23,101 @@ namespace smooth::core::io::spi
...
+                Log::error(log_tag, "HSPI_HOST Initialization failed");
             }
             else
             {
-                Log::verbose(log_tag, "SPI initialized, Host {}, DMA {}", host, dma_channel);
+                Log::info(log_tag, "HSPI initialized, Host {}, DMA {}", spi_host, dma_channel);
             }
         }
+
+        return hspi_initialized;
```
PerMalmberg, line 106: Do I remember correctly that we can only use either VSPI or HSPI, not both?

These two initializers are very similar and I don't see hspi_initialized or vspi_initialized being used other than to determine what to deinitialize. How about making them into a single function and instead passing hspi or vspi to it? For initialization, we only really need to save which bus is being used and if it is initialized or not.

lib/smooth/core/io/spi/Master.cpp
```diff
@@ -23,63 +23,101 @@ namespace smooth::core::io::spi
...
+            vspi_initialized = false;
+            spi_bus_free(VSPI_HOST);
+        }
+
+        if ((spi_host == HSPI_HOST) & (hspi_initialized))
         {
-            initialized = false;
-            spi_bus_free(host);
+            hspi_initialized = false;
+            spi_bus_free(HSPI_HOST);
```
PerMalmberg, line 120: With the previous comment in mind, this can be written as

if (initialized)
        {
            initialized = false;
            spi_bus_free(spi_host);
        }

CMakeLists.txt
```diff
@@ -22,7 +22,7 @@ set(CMAKE_CXX_STANDARD 17)
 
 
 # Select the test project to build
-set(selected_test_project starter_example)
+set(selected_test_project access_point)
```
PerMalmberg, line 25: Please keep the starter project as the default.

lib/smooth/core/io/spi/Master.cpp
```diff
@@ -23,63 +23,101 @@ namespace smooth::core::io::spi
...
+    bool Master::initialize(spi_host_device_t host,
+                            SPI_DMA_Channel dma_chl,
+                            gpio_num_t mosi,
+                            gpio_num_t miso,
+                            gpio_num_t clock,
+                            int transfer_size,
+                            gpio_num_t quadwp_io_num,
+                            gpio_num_t quadhd_io_num)
     {
+        std::lock_guard<std::mutex> lock(guard);
```
PerMalmberg, line 44: When we have the lock, shouldn't we check if we're already initialized before doing anything?

PerMalmberg (changes requested):
Nice to see it didn't require large changes to the existing code base.

lib/smooth/core/filesystem/MMCSDCard.cpp
```diff
@@ -67,6 +72,14 @@ namespace smooth::core::filesystem
         gpio_set_pull_mode(data2, GPIO_PULLUP_ONLY);   // D2, needed in 4-line mode only
         gpio_set_pull_mode(data3, GPIO_PULLUP_ONLY);   // D3, needed in 4- and 1-line modes
 
-        return do_common_initialization(mount_point, max_file_count, format_on_mount_failure, &slot_config);
+        auto mount_result = esp_vfs_fat_sdmmc_mount((*mount_point).c_str(),
+        &sdmmc_host,
+        &slot_config,
+        &mount_config,
+        &card);
```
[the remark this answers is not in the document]
enelson1001: The originally all the parameters were in a single line and uncrustify placed them at that position.  I went in and re-aligned and uncrustify left them as I placed them.

lib/smooth/core/filesystem/SDCard.cpp
enelson1001: Will add reference counters to Master one for hspi and one for vspi.

lib/smooth/core/io/spi/Master.cpp
enelson1001: No that is not correct both can be used.  I use vspi for lcd display & sdcard and hspi for connecting to SPI-BME280.  So I think we need both.  Plus this is were I will add the counter.

lib/smooth/core/io/spi/Master.cpp
enelson1001: See previous comment

CMakeLists.txt
enelson1001: OK

lib/smooth/core/io/spi/Master.cpp
enelson1001: I think now that we are keep count of the number of vspi_initialized and hspi_initialized we can't just check if either are initialized.

enelson1001: this is what Master.cpp looks like now - lots of duplicate code.
```
#include "smooth/core/io/spi/Master.h"
#include "smooth/core/logging/log.h"

using namespace smooth::core::logging;

namespace smooth::core::io::spi
{
    static constexpr const char* log_tag = "SPIMaster";

    // Declare static variables
    bool Master::hspi_initialized = false;
    bool Master::vspi_initialized = false;
    std::mutex Master::guard{};
    spi_bus_config_t Master::bus_config{};
    spi_host_device_t Master::spi_host;
    SPI_DMA_Channel Master::dma_channel;
    uint8_t Master::hspi_initialized_count = 0;
    uint8_t Master::vspi_initialized_count = 0;

    bool Master::initialize(spi_host_device_t host,
                            SPI_DMA_Channel dma_chl,
                            gpio_num_t mosi,
                            gpio_num_t miso,
                            gpio_num_t clock,
                            int transfer_size,
                            gpio_num_t quadwp_io_num,
                            gpio_num_t quadhd_io_num)
    {
        std::lock_guard<std::mutex> lock(guard);

        spi_host = host;
        dma_channel = dma_chl;

        bus_config.mosi_io_num = mosi;
        bus_config.miso_io_num = miso;
        bus_config.sclk_io_num = clock;
        bus_config.quadwp_io_num = quadwp_io_num;
        bus_config.quadhd_io_num = quadhd_io_num;
        bus_config.max_transfer_sz = transfer_size;

        bool initialized = false;

        if (spi_host == VSPI_HOST)
        {
            initialized = do_vspi_initialization();
        }

        if (spi_host == HSPI_HOST)
        {
            initialized = do_hspi_initialization();
        }

        return initialized;
    }

    bool Master::do_vspi_initialization()
    {
        if (!vspi_initialized)
        {
            vspi_initialized = spi_bus_initialize(spi_host, &bus_config, dma_channel) == ESP_OK;

            if (!vspi_initialized)
            {
                Log::error(log_tag, "VSPI_HOST Initialization failed");
            }
            else
            {
                Log::info(log_tag, "VSPI initialized, Host {}, DMA {}", spi_host, dma_channel);
            }
        }
        
        // increment the vspi_initialized_count so we know how many spi-devices are on the vspi-bus
        if (vspi_initialized)
        {
            vspi_initialized_count++;
        }

        return vspi_initialized;
    }

    bool Master::do_hspi_initialization()
    {
        if (!hspi_initialized)
        {
            hspi_initialized = spi_bus_initialize(spi_host, &bus_config, dma_channel) == ESP_OK;

            if (!hspi_initialized)
            {
                Log::error(log_tag, "HSPI_HOST Initialization failed");
            }
            else
            {
                Log::info(log_tag, "HSPI initialized, Host {}, DMA {}", spi_host, dma_channel);
            }
        }

        // increment the hspi_initialized_count so we know how many spi-devices are on the hspi-bus
        if (hspi_initialized)
        {
            hspi_initialized_count++;
        }

        return hspi_initialized;
    }

    void Master::deinitialize(spi_host_device_t spi_host)
    {
        std::lock_guard<std::mutex> lock(guard);

        if ((spi_host == VSPI_HOST) & vspi_initialized)
        {
            vspi_initialized_count--;

            if (vspi_initialized_count == 0)
            {
                vspi_initialized = false;
                spi_bus_free(VSPI_HOST);
            }
        } 
       

        if ((spi_host == HSPI_HOST) & hspi_initialized)
        {
            hspi_initialized_count--;

            if (hspi_initialized_count == 0)
            {
                hspi_initialized = false;
                spi_bus_free(VSPI_HOST);
            }
        } 
    }
}
```

enelson1001: Will supply updated code shortlu

lib/smooth/core/io/spi/Master.cpp
enelson1001: I tested newest code on ESP32-DevKit V1 with 2.8" TFT LCD with Touchscreen Breakout Board w/MicroSD Socket - ILI9341 (LCD and SDCard were on the VSPI_HOST bus) and the BME280-spi was placed on the HSPI_HOST bus.  Both spi-bus work.  The LCD display, SDCard and BME280 work without problems.  Also tested deinit the SDCard and the SDCard was unmounted and the initialized_count prevented the VSPI_HOST bus from being freed so the LCD display work.

lib/smooth/core/filesystem/SPISDCard.cpp
```diff
@@ -45,6 +45,7 @@ namespace smooth::core::filesystem
 
     bool SPISDCard::init(const SDCardMount& mount_point, bool format_on_mount_failure, int max_file_count)
     {
+        //card_is_spi_sd = true;
```
PerMalmberg, line 48: Remove :)

lib/smooth/core/io/spi/Master.cpp
```diff
@@ -53,71 +56,92 @@ namespace smooth::core::io::spi
...
-
-            if (!hspi_initialized)
-            {
-                Log::error(log_tag, "HSPI_HOST Initialization failed");
-            }
-            else
-            {
-                Log::info(log_tag, "HSPI initialized, Host {}, DMA {}", spi_host, dma_channel);
-            }
+            initialized_count++;
```
PerMalmberg, line 96: This member needs to be protected to support multi threading. Both init and de-init paths must ensure they lock the mutex so that only one path is run at a time.

CMakeLists.txt
PerMalmberg: This is still showing access_point as the selected project.

PerMalmberg (changes requested):

lib/smooth/core/io/spi/Master.cpp
enelson1001: Something is strange here.  Can you look at the actual file Master.cpp and Master.h because it looks like changes tracked by git are not correct.

It was probably something I did with git.  I followed theses steps https://www.tomasbeuzen.com/post/git-fork-branch-pull/
and got up to step 7 and opened this pull request.  When you requested changes I repeated step 6    $git push origin spi-master-updates but I'm not sure if this is correct.

enelson1001: Also CMakeLists.txt looks correct when I open the file Smooth/CMakeLists.txt in spi-master-updates.
So not sure what to do???

enelson1001: I will wait for your comment from my comments above before I correct SPISDCard.cpp just in case there are other changes.

PerMalmberg: It's OK, I must have missed it, the lock is there.

The selected project is however actually not correct, at least not what you have pushed to git:
https://github.com/enelson1001/Smooth/blob/spi-master-updates/CMakeLists.txt#L25

enelson1001: OK this last PUSH should complete your requests.  Hopefully next pull request I do will be better.

PerMalmberg (approved):

PR #150, opened 2021-03-27
Updates for new I2C sensor SHT30

enelson1001:
- Tested all 6 repeatability modes
-  Did minor cleanup DHT12.cpp
-  Just noticed CMakeLists.txt changed selected_test_project to be i2c_sht30_test so will immediately update and change back to starter_example.  I am not sure how this happened?
- The only other name I could think of for read_block was read_bytes.


PerMalmberg (commented):
Ok, so it wasn't possible to combine the existing write/read functions to achieve the desired functionality. If you say that's the case then I'll take your word for it.

Let me know if you're happy with this as-is and we'll get it merged.

enelson1001:
### I2C Stretching
I did more research on i2c and clock stretching and i2c-bus.org has very good explanation. https://www.i2c-bus.org/clock-stretching/
So with clock stretching enabled on the SHT30 it pulls down the SCL line after ACKing to i2c_read request.  The SHT30 will hold the SCL line low for a measurement duration (see table 4 and table 9). The different amounts of time depending upon which repeatability is chosen, so for Low Repeatability the worst case measurement time is 4.5ms, Medium Repeatability it is 6.5ms and High Repeatability it is 15.5ms (as shown in table 4).
### ESP32 I2C Stretching
A clock stretching question was asked on https://github.com/espressif/esp-idf/issues/4173 So it looks like the esp32 supports clock stretching but the maximum timeout is 13ms.
### What I did
So what I did was make a copy of your read function and called it read16 and implemented a 16bit slave register instead of an 8bit slave register. The read portion of the code failed.  So next I read what the master i2c_timeout was set to and found it was set to a count of 8000.  So then I changed the timeout to the maximum count of 1048575 (0xFFFFF) or approx. 13ms and then the new read16 worked.  But it is not the best fit for the SHT30 because the High Repeatability with clock stretching (measurement duration) can be 15.5ms.  I think I got lucky it worked on High Repeatability with clock stretching enabled because measurement time typically is 12.5ms and that is what I was seeing on my logic analyzer.
### Proposed changes to I2CMasterDevice.cpp
1. Keep read_block(uint8_t address, core::util::FixedBufferBase<uint8_t>& dest,  bool end_with_nack = true);
2. To support 16bit slave register, implement read16 function that supports a 16 bit register address based on your read function.
2. To support clock stretching
     -  Implement set_master_i2c_timeout function that uses i2c_set_timeout(i2c_port_t i2c_num, int timeout)
     -  implement get_master_i2c_timeout function that uses i2c_get_timeout(i2c_port_t i2c_num, int *timeout)
### Proposed changes to SHT30
1. Update my default Repeatability repeatability_command to  Repeatabilty::DisableMeasureHighRepeatabilityWithClockStretching  so it is not using clock stretching and use read_block.  
2. I will keep all 6 Repeatability modes in case someone would like to use for Medium/Low Repeatability with clock stretching and the new read16 function.
### Changes to Test Program i2c_sht30_test.cpp
1. I guess the test program could be changed to use both read_block and read16 functions??
### So what would you like to do
Let me know how you want to proceed.

PerMalmberg:
Sorry for the late reply. Since longer clock stretching isn't supported in the esp not using that sounds like the way to go. Maybe put a note next to the enum explaining why it is off by default.

enelson1001:
Tested above with SHT30 and DHT12
Created conditions to produce error messages.  They look like the following
```
E (36322) I2CMasterDevice: Error in - read16 - slave address: 0x44 - Operation timeout, bus busy

E (37383) I2CMasterDevice: Error in - read_block - slave address: 0x44 - Operation timeout, bus busy
```

lib/smooth/application/io/i2c/SHT30.cpp
```diff
@@ -228,7 +234,59 @@ namespace smooth::application::sensor
...
+        bool res = false;
+
+        if (use_clock_stretching)
+        {
+            res = execute_read16(raw_data);
+        }
+        else
+        {
+            res = execute_repeatability_command();
+            res &= execute_read_block(raw_data);
```
PerMalmberg, line 270: I know I'm guilty of using this construct too, but I've  since then learned that &= does *not* evaluate using short circuit-logic so please change it as follows.

```suggestion
            res = execute_repeatability_command() && execute_read_block(raw_data);
```

lib/smooth/application/io/i2c/SHT30.cpp
```diff
@@ -125,15 +117,29 @@ namespace smooth::application::sensor
...
-    void SHT30::set_repeatability_mode(Repeatabilty mode)
+    void SHT30::set_repeatability_mode(RepeatabilityMode mode)
     {
-        repeatability_command = mode;
+        // We will only use clock stretching on Medium and Low Repeatability Measurements
+        // since the ESP32 maximum i2c-master-timeout is 13ms.  If user selects
+        // EnableMeasureHighRepeatabilityWithClockStretching it will be implemented using
+        // MeasureHighRepeatabilityWithClockStretchingDisabled
+        if ((mode == RepeatabilityMode::EnableMeasureMediumRepeatabilityWithClockStretching)
+            | (mode == RepeatabilityMode::EnableMeasureLowRepeatabilityWithClockStretching))
```
PerMalmberg, line 127: Although this evaluates to the same, please use || instead, it prevents the implicit cast from int to bool.

```suggestion
        if (mode == RepeatabilityMode::EnableMeasureMediumRepeatabilityWithClockStretching
            || mode == RepeatabilityMode::EnableMeasureLowRepeatabilityWithClockStretching)
```

lib/smooth/application/io/i2c/SHT30.cpp
```diff
@@ -228,7 +234,59 @@ namespace smooth::application::sensor
...
+        }
+
+        return res;
+    }
+
+    // Execute status read
+    bool SHT30::execute_status_read(FixedBufferBase<uint8_t>& raw_data)
+    {
+        auto res = execute_write_command(Command::ReadStatus);
+        res &= execute_read_block(raw_data);
```
PerMalmberg, line 285:
```suggestion
        auto res = execute_write_command(Command::ReadStatus) && execute_read_block(raw_data);
```

lib/smooth/core/io/i2c/I2CMasterDevice.cpp
```diff
@@ -24,77 +24,148 @@ limitations under the License.
...
         // Set R/W bit to 0 for write.
-        address = static_cast<uint8_t>(address << 1);
+        auto write_address = static_cast<uint8_t>(address << 1);
 
         auto res = i2c_master_start(link);
-        res |= i2c_master_write_byte(link, address, expect_ack);
+        res |= i2c_master_write_byte(link, write_address, expect_ack);
         res |= i2c_master_write(link, data.data(), data.size(), expect_ack);
         res |= i2c_master_stop(link);
+        res |= i2c_master_cmd_begin(port, link, to_tick(timeout));
```
PerMalmberg, line 57: Funny how I've not seen this before. These should have been `&=`, but with the new learning as noted in another comment, we need to use &&.

```suggestion
auto res = i2c_master_start(link)
        && i2c_master_write_byte(link, write_address, expect_ack);
        && i2c_master_write(link, data.data(), data.size(), expect_ack);
        && i2c_master_stop(link);
        && i2c_master_cmd_begin(port, link, to_tick(timeout));
```

lib/smooth/core/io/i2c/I2CMasterDevice.cpp
```diff
@@ -24,77 +24,148 @@ limitations under the License.
...
 
+        // Set the process name that will be used by log_error
+        set_process_name(read16_tag);
+
+        // Convert slave register to FixedBuffer
+        FixedBuffer<uint8_t, 2> slave_reg;
+        slave_reg[0] = static_cast<uint8_t>(slave_register >> 8);
+        slave_reg[1] = static_cast<uint8_t>(slave_register & 0x00FF);
+
+        return write_followed_by_read(link, address, slave_reg, data, use_restart_signal, end_with_nack, scl_timeout);
```
PerMalmberg, line 131: As already noted, |= is not the correct operator to use for the chained calls throughout this function. If the first call fails, we shouldn't continue the operation.

lib/smooth/core/io/i2c/I2CMasterDevice.cpp
```diff
@@ -126,26 +197,36 @@ namespace smooth::core::io::i2c
...
+            do_cleanup();
         }
 
         return res == ESP_OK;
     }
 
+    // Write read address and read data
     bool I2CMasterDevice::read_block(uint8_t address,
-                                     core::util::FixedBufferBase<uint8_t>& data,
+                                     FixedBufferBase<uint8_t>& data,
```
PerMalmberg, line 221: Same here, && instead of |=. Sorry you'll have to clean up my mistake in this PR :)

lib/smooth/core/io/i2c/I2CMasterDevice.cpp
```diff
@@ -217,27 +299,45 @@ namespace smooth::core::io::i2c
...
+    //void I2CMasterDevice::log_error(esp_err_t err, const char* msg)
+    void I2CMasterDevice::log_error(esp_err_t err, uint8_t address)
     {
+        std::stringstream ss;
+        ss << "Error in - " << process_name << " - slave address: 0x" << std::hex << static_cast<int32_t>(address);
+
         if (err == ESP_ERR_INVALID_ARG)
         {
-            Log::error(log_tag, "{} - Parameter error", msg);
+            Log::error(log_tag, "{} - Parameter error", ss.str().c_str());
```
PerMalmberg, line 311: Is c_str() really needed, doesn't fmt accept a string as an argument?

lib/smooth/include/smooth/core/io/i2c/I2CMasterDevice.h
```diff
@@ -74,7 +74,8 @@ namespace smooth::core::io::i2c
             /// \return true on success, false on failure
             bool write(uint8_t address, std::vector<uint8_t>& data, bool expect_ack = true);
 
-            /// Reads data from the register of the slave with the provided address.
+            /// Reads data from the 8 bit register of the slave with the provided address.  This function DOES NOT
+            // support slave clock-stretching
```
PerMalmberg, line 78:
```suggestion
            /// support slave clock-stretching
```

lib/smooth/include/smooth/core/io/i2c/I2CMasterDevice.h
```diff
@@ -88,7 +89,46 @@ namespace smooth::core::io::i2c
...
                       bool end_with_nack = true);
 
-            /// Reads a block of data from the slave with the provided address.
+            /// Reads data from the 8 bit register of the slave with the provided address. This function supports slave
+            // clock-stretching
+            /// \param address The slave address
+            /// \param slave_register The 16 bit register to read from.
+            /// \param dest Where the data will be written to. The size of the buffer determines how many bytes to read.
+            /// \param use_restart_signal If true, uses a start-condition instead of a stop-condition after the slave
+            // address.
```
PerMalmberg, line 98:
```suggestion
            /// address.
```

lib/smooth/include/smooth/core/io/i2c/I2CMasterDevice.h
```diff
@@ -88,7 +89,46 @@ namespace smooth::core::io::i2c
...
+                       bool end_with_nack = true,
+                       int scl_timeout = 0);
+
+            /// Reads data from the 16 bit register of the slave with the provided address. This function supports slave
+            // clock-stretching
+            /// \param address The slave address
+            /// \param slave_register The 16 bit register to read from.
+            /// \param dest Where the data will be written to. The size of the buffer determines how many bytes to read.
+            /// \param use_restart_signal If true, uses a start-condition instead of a stop-condition after the slave
+            // address.
```
PerMalmberg, line 117:
```suggestion
            /// address.
```

lib/smooth/include/smooth/core/io/i2c/I2CMasterDevice.h
```diff
@@ -101,7 +141,38 @@ namespace smooth::core::io::i2c
...
+            /// Write followed by read
+            /// \param link The I2C link command
+            /// \param address The slave address
+            /// \param slave_reg A FixedBuffer that hold the slave register address
+            /// \param dest Where the data will be written to. The size of the buffer determines how many bytes to read.
+            /// \param use_restart_signal If true, uses a start-condition instead of a stop-condition after the slave
+            // address.
+            /// \param end_with_nack If true, ends the transmission with a NACK instead of an ACK.
+            /// \param scl_timeout Used to suppport slave clock-stretching.  Slave must release SCL line within this
+            // time.  Set to 0 to disable using slave clock-stretching.  The maximum timeout value is 0xFFFFF
```
PerMalmberg, line 153:
```suggestion
            /// time.  Set to 0 to disable using slave clock-stretching.  The maximum timeout value is 0xFFFFF
```

lib/smooth/include/smooth/core/io/i2c/I2CMasterDevice.h
```diff
@@ -101,7 +141,38 @@ namespace smooth::core::io::i2c
...
             i2c_port_t port;
         private:
-            void log_error(esp_err_t err, const char* msg);
+            /// Write followed by read
+            /// \param link The I2C link command
+            /// \param address The slave address
+            /// \param slave_reg A FixedBuffer that hold the slave register address
+            /// \param dest Where the data will be written to. The size of the buffer determines how many bytes to read.
+            /// \param use_restart_signal If true, uses a start-condition instead of a stop-condition after the slave
+            // address.
```
PerMalmberg, line 150:
```suggestion
            /// address.
```

lib/smooth/include/smooth/core/io/i2c/I2CMasterDevice.h
```diff
@@ -101,7 +141,38 @@ namespace smooth::core::io::i2c
...
+            /// \param end_with_nack If true, ends the transmission with a NACK instead of an ACK.
+            /// \param scl_timeout Used to suppport slave clock-stretching.  Slave must release SCL line within this
+            // time.  Set to 0 to disable using slave clock-stretching.  The maximum timeout value is 0xFFFFF
+            // (1,048,575) or 13ms.  Negative numbers will be ignored.
```
PerMalmberg, line 154:
```suggestion
            /// (1,048,575) or 13ms.  Negative numbers will be ignored.
```

PerMalmberg (changes requested):
Also don't forget to update the contribution file.

lib/smooth/include/smooth/core/io/i2c/I2CMasterDevice.h
enelson1001: This is how uncrustify formats the document, but will change.  I think it looks better with three forward slashes then 2 forward slashes

lib/smooth/core/io/i2c/I2CMasterDevice.cpp
enelson1001: Yes you are correct - removing c_str()

enelson1001:
I tried changing res operations to &= or && but log_error was never called because the resultant of these operations was always ESP_OK because ESP_OK=0.  So changed back |= and | so it looks like your original code was correct.  I'm not sure but I think the command link stops execution if one of the commands does not equal ESP_OK otherwise if we have multiple failures and start ORing all the results we could have a weird value at the end.

I think I got all your requested updates.
I'll updated contribution file once you say everything looks good - I guess I need to add #146 and #150

lib/smooth/core/io/i2c/I2CMasterDevice.cpp
```diff
@@ -154,17 +157,14 @@ namespace smooth::core::io::i2c
         // before changing so we can restore before returning from this function.
         if (scl_timeout > 0)
         {
+            //Log::info("I2C", "timeout = {}", scl_timeout);
```
PerMalmberg, line 160:
```suggestion
```

PerMalmberg (changes requested):

PerMalmberg:
Oh, yeah. The joy of having OK = 0.... oh well. Sometime we might hcnage it to call() == ESP_OK instead.

I looks good otherwise.

PerMalmberg (changes requested):

lib/smooth/core/io/i2c/I2CMasterDevice.cpp
```diff
@@ -157,13 +156,17 @@ namespace smooth::core::io::i2c
...
         {
-            //Log::info("I2C", "timeout = {}", scl_timeout);
-            res |= i2c_get_timeout(port, &orig_scl_timeout);
-            res |= i2c_set_timeout(port, scl_timeout);
+            auto timeout_res = i2c_get_timeout(port, &orig_scl_timeout);
+            timeout_res |= i2c_set_timeout(port, scl_timeout);
+
+            if (timeout_res != ESP_OK)
+            {
+                log_error(timeout_res, address);
```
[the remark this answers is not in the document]
enelson1001: Yes I think that is a good idea.  Will implement

lib/smooth/core/io/i2c/I2CMasterDevice.cpp
```diff
@@ -204,7 +207,12 @@ namespace smooth::core::io::i2c
         // If we implemented slave clock-stretching then restore master timeout to its original value
         if (scl_timeout > 0)
         {
-            res |= i2c_set_timeout(port, orig_scl_timeout);
+            auto timeout_res = i2c_set_timeout(port, orig_scl_timeout);
+
+            if (timeout_res != ESP_OK)
+            {
+                log_error(timeout_res, address);
```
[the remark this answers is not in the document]
enelson1001: I think I will remove the error checking here since if the read, write above failed we will exit and if it didn't fail then this write should succeed.  What do you think?

enelson1001:
I've updated to latest master (PR #151) but now have problems running CI/.build_test.sh on local PC.  Errors out when trying to build test programs.
```
Building esp32 test projects -----------------------------------------
/src/CI/container_scripts/prepare_idf.sh: line 3: .: filename argument required
.: usage: . filename [arguments]
/esp/esp-idf /src
Git describe tags -----------------------------------------------------
v4.4-dev-4-g73db14240
/src
#######################################
Compiling project access_point
#######################################

-- Compiling for ESP
-- Found Git: /usr/bin/git (found version "2.27.0") 
-- IDF_TARGET not set, using default target: esp32
-- The C compiler identification is unknown
-- The CXX compiler identification is unknown
-- The ASM compiler identification is unknown
-- Found assembler: xtensa-esp32-elf-gcc
CMake Error at /esp/esp-idf/tools/cmake/project.cmake:308 (__project):
  The CMAKE_C_COMPILER:

    xtensa-esp32-elf-gcc

  is not a full path and was not found in the PATH.

  Tell CMake where to find the compiler by setting either the environment
  variable "CC" or the CMake cache entry CMAKE_C_COMPILER to the full path to
  the compiler, or to the compiler name if it is in the PATH.
Call Stack (most recent call first):
  CMakeLists.txt:74 (project)


CMake Error at /esp/esp-idf/tools/cmake/project.cmake:308 (__project):
  The CMAKE_CXX_COMPILER:

    xtensa-esp32-elf-g++

  is not a full path and was not found in the PATH.

  Tell CMake where to find the compiler by setting either the environment
  variable "CXX" or the CMake cache entry CMAKE_CXX_COMPILER to the full path
  to the compiler, or to the compiler name if it is in the PATH.
Call Stack (most recent call first):
  CMakeLists.txt:74 (project)


CMake Error at /esp/esp-idf/tools/cmake/project.cmake:308 (__project):
  The CMAKE_ASM_COMPILER:

    xtensa-esp32-elf-gcc

  is not a full path and was not found in the PATH.

  Tell CMake where to find the compiler by setting either the environment
  variable "ASM" or the CMake cache entry CMAKE_ASM_COMPILER to the full path
  to the compiler, or to the compiler name if it is in the PATH.
Call Stack (most recent call first):
  CMakeLists.txt:74 (project)


-- Warning: Did not find file Compiler/-ASM
-- Configuring incomplete, errors occurred!
See also "/src/build/CMakeFiles/CMakeOutput.log".
See also "/src/build/CMakeFiles/CMakeError.log".
```

lib/smooth/core/io/i2c/I2CMasterDevice.cpp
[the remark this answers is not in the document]
PerMalmberg: No, keep it. There's no guarantee that an operation succeeds just because the one prior to it did. You never know when there is interference.

PerMalmberg:
`/src/CI/container_scripts/prepare_idf.sh: line 3: .: filename argument required` must be this line `. $IDF_TOOLS_EXPORT_CMD` so must be empty for you for some reason. 

I can't figure out why and I'm sure I ran the tests from latest master after merging that PR.

enelson1001:
I ended up changing **. $IDF_TOOLS_EXPORT_CMD**    to this ----->   **. $IDF_PATH/export.sh** and it worked for me.  Are you OK with me submitting this change?

PerMalmberg:
> I ended up changing **. $IDF_TOOLS_EXPORT_CMD** to this -----> **. $IDF_PATH/export.sh** and it worked for me. Are you OK with me submitting this change?

Let's hear if @peterus has some input as to why IDF_TOOLS_EXPORT_CMD would be empty.

peterus:
it looks like @enelson1001 is not using the newest docker image.
please delete the local docker image with this commands:
`docker image rm permalmberg/smooth`
and run the build command again.

@PerMalmberg maybe we should add some kind of version information (as the docker image tag) to the docker image.
This can/should be based on the lastest idf version tag. using latest is not always the best as you can see in this example.

PerMalmberg:
> it looks like @enelson1001 is not using the newest docker image.
> please delete the local docker image with this commands:
> `docker image rm permalmberg/smooth`
> and run the build command again.
> 
> @PerMalmberg maybe we should add some kind of version information (as the docker image tag) to the docker image.
> This can/should be based on the lastest idf version tag. using latest is not always the best as you can see in this example.

Yes, that's a good idea.

enelson1001:
deleting the local docker image did the trick and build_test.sh ran with no error.

PerMalmberg (approved):

Issue #152, opened 2021-04-03
github actions: compile host and esp32 build in parallel


peterus:
I suggest to compile the host- and esp32-build in parallel to reduce runtime.

the `CI`-job (https://github.com/PerMalmberg/Smooth/blob/master/.github/workflows/ci.yml) should be splitted up in `CI-host` and `CI-esp32` jobs. the corresponding scripts should be started in the different jobs.
This split will reduce the runtime dramatically (would expect the runtime will be just the half).

If this change is okay for you, I am happy to do the changes.

PerMalmberg:
Absolutely. please do. If possible, I'd like to see a script that runs both, for use locally.

peterus:
Branch merged - closing Ticket

PR #156, opened 2021-04-09
build in parallel and some cleanup

peterus:
I did some small tasks in this pull request:

- parallel build of host / esp32 (closing #152)
- add docker-compose which is the 'standard' tool nowadays if you have repeated setups
- simpler build scripts


.github/workflows/ci.yml
```diff
@@ -1,23 +1,32 @@
...
-    - name: Get submodules
-      run: git submodule update --init --checkout --recursive
-    - name: Build and test
-      run: |
-        ./CI/build_test.sh
+    - uses: actions/checkout@v2
+      with: 
+        submodules: true
+    - name: Build
+      run: docker-compose run --rm smooth ./CI//build_smooth_host.sh
```
PerMalmberg, line 13:
```suggestion
      run: docker-compose run --rm smooth ./CI/build_smooth_host.sh
```

CI/build_smooth_host.sh
```diff
@@ -0,0 +1,44 @@
...
+
+    sed -i -r -e "$pattern" ../CMakeLists.txt
+
+    cmake .. -G "Ninja" -DCMAKE_TOOLCHAIN_FILE=../CI/gnu-toolchain.cmake
+    ninja
+
+done
+
+# Clean up
+cp ./original_CMakeLists.txt ../CMakeLists.txt
```
PerMalmberg, line 44: This is just a copy of build_smooth_eps.sh, isn't it? Can we use a single file for both? Unless they diverge somehow I don't see a reason to have two copies of the same script.

PerMalmberg (changes requested):

PerMalmberg:
Should we take the opportunity to also set a specific IDF-version as discussed in #150 in this PR or shall we do that in a separate PR?

I'm thinking that using a .env file with the IDF-version would be the best way. That'd allow for a single place to set the IDF-version from which both `docker build` and `docker-compose` can pick it up, both when run locally and and in CI, assuming the file can be placed in a location found by both. 

By updating docker.compose.yml, Dockerfile and build-docker-image.sh we could even allow easy switching between versions locally if someone wants to run against another version than what we set as default (v4.2?)

CI/build_smooth_host.sh
peterus: in the ende you are right, but there are still some differences:
- the esp version is not compiling the linux tests
- the build directory and the toolchain is different

we could merge all this into one script file, but then we need an input argument. I am open for both solutions, as you like it.

peterus:
from an other thread I was reading that you like building against the latest master branch. So I was not considering this solution.
But if you like to build against a dedicated IDF version I can definitely do this with the `.env` file. I like your idea with the `.env` file, with this everything is in one place.

CI/build_smooth_host.sh
PerMalmberg: Right. Lets leave it for now and do something better when/if they these files needs to be updated.

PerMalmberg:
I do like building against the master branch, but now that Espressif (allegedly) have fixed the PSRAM issues that were only available in master, that need has lessened. And, with the sudden upsurge in users of Smooth, it makes sense to build against an official release of IDF, while still giving the option to build against master by simply changing a file and building the image locally.

peterus:
as you can see now, with the `.env` file it is very easy to switch versions. also changing the docker image name is now simple.

If it is okay for you I would try to implement some kind of check if there is a newer IDF image available, building the new image and uploading it to your docker hub or github packages if all tests are successful. of course i would need some kind of API key in the github actions environment.

PerMalmberg:
Very nice.

I'm not sure automatically updating the image is the right way to go as it introduces a moving target in he build chain. However, a step that fails if there is a new release available could be a good thing if it isn't set to be a required success in the branch protection settings.

I like the idea of building against both `latest` and the latest major branch of IDF.

peterus:
I had an idea which I want to present to you:
- I splited up the image into a host and esp32 image. As the host image is very small and building up very fast (nearly the same time as docker has to download the image), it will be build as it will be needed. with this you do not have to publish a dedicated docker image.
- with the help of github actions the CI can run against various IDF-versions (I tried 4.0 until 4.3 and the master branch, IDF-version 4.0 and 4.1 are failing because of interface issues (which is expected)).

I like this approach very much as it is very clean and easy to maintain.

PerMalmberg:
Oh, that's a really nice solution. Once Ubuntu 20.04 is available we can update the script to that.

peterus:
Ubuntu 20.04 is already available in docker :)
I updated the Dockerfile

CI/host-docker/Dockerfile
```diff
@@ -1,8 +1,14 @@
-FROM espressif/idf
+
+#ARG  ESP_IDF_VERSION=latest
```
PerMalmberg, line 2:
```suggestion
```

.env
```diff
@@ -0,0 +1 @@
+ESP_IDF_VERSION=release-v4.2
```
PerMalmberg, line 1:
```suggestion
# Change this to the ESP-IDF version you want to run local CI tests against when running the CI/build_test.sh script
ESP_IDF_VERSION=release-v4.2
```

.github/workflows/ci.yml
```diff
@@ -1,23 +1,40 @@
...
+    - name: Build
+      run: docker-compose run --rm host ./CI/build_smooth_host.sh
+    - name: Test
+      run: docker-compose run --rm -w /src/build_host/test/linux_unit_tests host ./linux_unit_tests
+  
+  CI_esp:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        esp_idf_version:
```
PerMalmberg, line 21: Just to keep variable names different for less confusion.
```suggestion
        test_against_idf_versions:
```

.github/workflows/ci.yml
```diff
@@ -1,23 +1,40 @@
...
+          - 'release-v4.3'
+          - 'release-v4.2'
+    steps:
+    - uses: actions/checkout@v2
+      with: 
+        submodules: true
+    - name: Build
+      run: docker-compose run --rm esp32 ./CI/build_smooth_esp32.sh
+      env:
+          ESP_IDF_VERSION: ${{ matrix.esp_idf_version }}
```
PerMalmberg, line 32:
```suggestion
          ESP_IDF_VERSION: ${{ matrix.test_against_idf_versions }}
```

PerMalmberg (changes requested):
Maybe a note in the readme explaining the build process, how to select an IDF-version locally?

Also, please update the contributors file to list this PR

peterus:
From my side I am now happy with the current changes.
If you still miss something, I would be happy to hear them.

README.md
```diff
@@ -26,7 +26,8 @@ with the only consideration that the mocks do not actually simulate the hardware
 * ESP-IDF v4.x
 * GCC 8
 
-Smooth is developed on a Linux machine so how well it compiles using the Windows toolset povided by Espressif is unknown. 
+Smooth is developed on a Linux machine so how well it compiles using the Windows toolset povided by Espressif is unknown.
+If you are working on Windows or you don't want to install the dependencies on your lokal machine you can use the docker images which are provided.
```
PerMalmberg, line 30:
```suggestion
If you are working on Windows or you don't want to install the dependencies on your local machine you can use the docker images which are provided.
```

README.md
```diff
@@ -267,3 +268,27 @@ int main(int /*argc*/, char** /*argv*/)
 
 }
 ```
+
+## Running CI scripts lokaly
```
PerMalmberg, line 272:
```suggestion
## Running CI scripts locally
```

README.md
```diff
@@ -267,3 +268,27 @@ int main(int /*argc*/, char** /*argv*/)
...
+
+If you want to test your changes in Smooth, you need to pass the CI on Github. To test your changes on your lokal system you can use docker:
```
PerMalmberg, line 274:
```suggestion
If you want to test your changes in Smooth, you need to pass the CI on Github. To test your changes on your local system you can use docker:
```

README.md
```diff
@@ -267,3 +268,27 @@ int main(int /*argc*/, char** /*argv*/)
...
+
+- to compile the host binaries: `docker-compose run --rm host ./CI/build_smooth_host.sh`
+- to run the host unit test: `docker-compose run --rm -w /src/build/host/test/linux_unit_tests host ./linux_unit_tests`
+- to compile the esp32 binaries: `docker-compose run --rm esp32 ./CI/build_smooth_esp32.sh`
+
+To run all this commands one by one you can run this script: `./CI/build_test.sh`
```
PerMalmberg, line 280:
```suggestion
To run these commands at once you can run this script: `./CI/build_test.sh`
```

README.md
```diff
@@ -267,3 +268,27 @@ int main(int /*argc*/, char** /*argv*/)
...
+
+### Run a host build with TCP ports
+
+On default `docker-compose` is not opening the ports in in the `run` mode. To open the TCP ports for server testing you need to enter the docker image in this way:
+
+`docker-compose run --service-ports --rm host`
+
+### choos different release branches for the ESP32 build
```
PerMalmberg, line 288:
```suggestion
### Choose different release branches for the ESP32 build
```

README.md
```diff
@@ -267,3 +268,27 @@ int main(int /*argc*/, char** /*argv*/)
...
+
+For the esp32 binaries the mainline branch is used on default. If you want to use a release branch you have to set the environment variable `ESP_IDF_VERSION`.
+Here is an example with the v4.2 release branch:
+`ESP_IDF_VERSION=release-v4.2 docker-compose run --rm esp32 ./CI/build_smooth_esp32.sh`
+
+All release branches are used on the Github CI which are compatible with Smooth for testing.
```
PerMalmberg, line 294:
```suggestion
In CI, all compatible branches of IDF are checked on each push and pull request.
```

PerMalmberg (changes requested):
Just a few typos to fix before merging.

peterus:
Ah stupid mistakes. I fixed them all.

PerMalmberg:
Awesome, merged!

PR #158, opened 2021-04-15
vscode updates

peterus:
most of the paths in the vsc settings where hard coded, i change the paths to relative paths and used the IDF variables where possible.


.vscode/c_cpp_properties.json
```diff
@@ -3,31 +3,31 @@
...
             "name": "host",
             "includePath": [
-                "/workspaces/Smooth/lib/smooth/include/**",
-                "/workspaces/Smooth/externals/**",
-                "/esp/esp-idf/components/**",
-                "/esp/tools/tools/xtensa-esp32-elf/esp-2020r1-8.2.0/xtensa-esp32-elf/xtensa-esp32-elf/include/**"
+                "lib/smooth/include/**",
+                "externals/**",
+                "mock-idf/components/**",
+                "$IDF_TOOLS_PATH/tools/xtensa-esp32-elf/esp-2020r3-8.4.0/xtensa-esp32-elf/include/**"
```
PerMalmberg, line 30: Might not be possible, but can this be made version agnostic by use of another variable?

PerMalmberg:
Oh, using the same image for dev as for CI. Neat.

peterus:
was thinking the same first 🙂
but than i started to play around and it is not really practical in development as you need to switch all the times container if you want to try it on host and esp at the same time. I am currently trying different solutions, I will keep you updated 😉

peterus:
I did some quick updates:
- i cleaned up the vscode settings, if i removed to much, just tell me 🙂
- i created a 3. docker image (dev) which is based on the idf and all missing things for the host build. with this one you can compile for esp and host at the same time.

what do you think about the 3 docker images? should we drop the esp and host one and go just for the dev image?

Will do more updates today on the evening.

PerMalmberg:
Cool, I see it was possible to make the path version agnostic.

Well, having a single image is convenient locally, but is there a benefit to have them separate in CI? If not then I see no reason to have all three.

peterus:
I changed it back to one image as it is just simpler for development. the only benefit would be for the CI as it would be a little bit faster, but in the end it will be just some seconds.

One additional question: are you using the "Remote - Containers" extension in VSCode? I am personally developing in the cloud as my machines at home are not powerful enough anymore and the extension is just working if it is running locally and not remote via ssh.
It is working now - but will it be needed?

PerMalmberg:
I've mostly used a native Linux machine, but next time I do anything actively with Smooth I'm probably using "Remote - Containers" in VS Code.

How do you hook up USB to your remote machines?

.devcontainer/docker-compose.yml
```diff
@@ -0,0 +1,7 @@
+version: "3.9"
+services:
+  smooth:
+    extends:
+      file: ../docker-compose.yml
+      service: smooth
+    command: /bin/sh -c "rm -rf /home/dev/.vscode-server && while sleep 1000; do :; done"
```
PerMalmberg, line 7: Why is this needed and what's in the /home/dev folder?

PerMalmberg (changes requested):

PerMalmberg:
@peterus Shall we try to complete this one? Only a single question that needs answers.

peterus:
@PerMalmberg sorry, i am very busy lately.
I will come back with a proper answer in some hours when the kid is sleeping 😊

if you want to take a look before hand: google "docker device". On linux you can forward the device (/dev/ttyUSB0) directly to the docker container. On Windows an Mac you need to use virtual box to work around some issues.

PerMalmberg:
No worries, just don't want to forget about these changes :)

.devcontainer/docker-compose.yml
peterus: this was a copy-past thing, i removed it.
the rest is needed to keep the docker container running.

peterus:
regarding hooking up the usb: you can use the device option: https://docs.docker.com/compose/compose-file/compose-file-v3/#devices

linux has a native supported for this, you can start the docker container just with the new option in the docker-compose file.

on windows and mac this is not so easy possible:
- on windows you don't have the "everything is a file" concept
- on mac docker is running in a special mode which is not allowing to forward devices.

but there is a workaround for it: use virtual box as a host machine and forward the usb ports to it.
howto: https://christopherjmcclellan.wordpress.com/2019/04/21/using-usb-with-docker-for-mac/

PerMalmberg (approved):

PerMalmberg:
Merged. :)

PR #163, opened 2021-07-17
Add new mono display

enelson1001:
added SSD1306 I2C version
fix build_test.sh
update README.md


PerMalmberg:
@enelson1001 Hey. I'm on vacation for another week, I'll have look at this afterwards.

PerMalmberg:
Looks good, just wonder if all the changes in sdkonfig are just default values for a newer IDF?

enelson1001:
Must be because I did not make changes to that  file.

README.md
```diff
@@ -64,12 +64,14 @@ If you are working on Windows or you don't want to install the dependencies on y
       - ILI9341
       - ST7735
       - SH1107
+      - SSD1306(I2C)
```
PerMalmberg, line 67: If this is a I2C device, then lets place it under the I2C section. Maybe we should reorganize the entire tree based on type instead of protocol type....

PerMalmberg (changes requested):

README.md
enelson1001: Are you saying do something like this
spi 
 |- display
 | - sensors

i2c
 |- display
 | sensors


etc?

PerMalmberg:
Yes, don't you think that is more organised that the current list?

enelson1001:
Yes that does sound better.  
1.Do you think we should include rmt as one of the types and move RGBled under it?  
2. Where would Weigand go?  
3. Still keep tree with io and under it spi, i2c and rmt?  
4. Under spi i2c rmt do we want additional folders like display, sensor, digital, analog, rtc, power mgt etc or just limp everything together?

PerMalmberg:
1. Sure, why not.
2. It's own category, maybe? "Oddballs" :)
3. If it adds something
4. If it doesn't just clutter the list, sure.

enelson1001:
I will start looking at what is involved to make this change this weekend.  I will probably have a few more comments.

enelson1001:
Per
Please comment on the attached proposal before I begin working on this.
[PullReq163_ProposedChanges.pdf](https://github.com/PerMalmberg/Smooth/files/7077606/PullReq163_ProposedChanges.pdf)

thanks

PR #166, opened 2021-08-04
Flag Sntp as started after starting

jsmestad:
This appears to be missing. I can't find a place where `started` is set to anything but the initial value of `started = false`


PerMalmberg:
Thanks. Can you please update the contributors file too, and correct the formatting to pass the CI-pipeline?

PR #170, opened 2022-05-06
Quick change to allow Smooth to compile on ESP32-S3.

tmiw:
Per https://github.com/espressif/esp-idf/blob/master/components/hal/include/hal/spi_types.h, HSPI_HOST and VSPI_HOST don't exist on ESP32-S3, causing Smooth to fail while building. This change removes all references to HSPI_HOST and VSPI_HOST in favor of SPI2_HOST and SPI3_HOST.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -19,7 +19,6 @@ jobs:
     strategy:
       matrix:
         test_against_idf_versions:
-          - 'release-v5.0-dev'
           - 'release-v4.4'
           - 'release-v4.3'
           - 'release-v4.2'
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -19,7 +19,8 @@ jobs:
     strategy:
       matrix:
         test_against_idf_versions:
-          - 'latest'
+          - 'release-v5.0-dev'
+          - 'release-v4.4'
           - 'release-v4.3'
           - 'release-v4.2'
     steps:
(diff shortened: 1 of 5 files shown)


PerMalmberg:
Thanks for the contribution, @tmiw  One of the tests are failing, seems libsodium is no longer available? I've not worked with the ESP for a long while now so I'm not up to speed what has changed in the esp-idf project.

tmiw:
> Thanks for the contribution, @tmiw One of the tests are failing, seems libsodium is no longer available? I've not worked with the ESP for a long while now so I'm not up to speed what has changed in the esp-idf project.

ESP-IDF isn't recognizing libsodium but the package does still exist on Ubuntu. I checked in a tweak to the CI to only use the tags listed in https://github.com/espressif/esp-idf/tags so we'll see how that goes.

tmiw:
> Wasn't libsodium part of esp-idf? A version ported to work with the esp?

Not sure. Since it was failing on "latest" before, perhaps it's some bug on Expressif's end.

In any case, the CI builds are using 4.2-4.4 now, which seem to be the official releases according to their Git. Those build fine FWIW.

PerMalmberg:
I removed the requirement in CI for it to build against the latest ESP, I don't have the time to maintain that myself and since it builds against the latest official tags it'll have to do for now.

PerMalmberg:
@tmiw Please add yourself to the contributor file and we'll merge this.

tmiw:
> @tmiw Please add yourself to the contributor file and we'll merge this.

Done.

